PR #31, opened 2023-01-04
Different address for remote access

Nikorag:
Currently, the automation only works with a single IP address. This change would allow you to define your home SSID and then a separate IP Address (or DNS entry) to use when out of the house.

Have tried the resultant script when on my home network and then tethered to my phone using the DNS entry (ensuring all the ports are open) and it works for me.

### DIFF
--- a/scripts/gen-launcher.sh
+++ b/scripts/gen-launcher.sh
@@ -89,8 +89,35 @@ do
     fi
 done
 
+PS3="Do you have a separate address (DNS or IP) to access this console away from home: "
+select away in "Yes" "No"
+do
+    if [ -z "${away}" ]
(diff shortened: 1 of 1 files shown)


finder39:
This would be great! I currently used automation to create the launcher and then modified the config file to replace all local ip addresses with the domain I use to access my house remotely (with the correct ports forwarded). Being able to have the automation work directly without having to take those steps would help a ton of people

streetpea:
Thanks for the pull @Nikorag. I'll merge a version of this with the additional hostname option and space matching that I added locally.

streetpea:
This has been merged as part of release v1.3.0

PR #60, opened 2023-04-06
Refactor color correction fix 

jonibim:
While the old solution by [Egoistically](https://github.com/Egoistically/chiaki/commit/0f12a639af52e2d294c7d1e22bd504bcb5f7fe8c) was sound, the code was not written in the same manner as in the original repository. The author also confirmed in [a commit](https://github.com/Egoistically/chiaki/commit/0f12a639af52e2d294c7d1e22bd504bcb5f7fe8c#commitcomment-87967719) that it was written quickly from another source. This PR respects the same coding style as the original repo.

In addition, the solution provided by the original author was not implemented for the hardware accelerated options. Therefore as soon as a streams is switched to make use of such an option, it will revert to the old incorrect colors. 

Based on this [comment](https://github.com/Egoistically/chiaki/commit/0f12a639af52e2d294c7d1e22bd504bcb5f7fe8c#commitcomment-87845721), I experimented with different color spaces since I believe the scaling matrix is adjusted for Full Range. Indeed that as soon as you adjust the scaling matrix for the BT701 Limited Range color space, you end up with the correct colors in the stream using any hardware acceleration setting. This fix is present on this PR.

A similar PR was also created on [Egoistically's fork](https://github.com/Egoistically/chiaki/pull/1) which was approved and merged.

P.S: I have only tested this with a PS4 since that is the only console I own


streetpea:
Thanks for the pull request. Sorry it took so long for me to get to it. I'll test this out with a ps5 and look to merge it.

Issue #93, opened 2023-11-13
HDR support


streetpea:
Add support for HDR image rendering in opengl for screens that support it such as the new Steam Deck OLED HDR screen.

jbaiter:
I'm thinking of picking this up, got briefed by Florian Grill on how to enable HDR and will try to work out how to pass it onto the OLED deck's screen, which is going to take some detective work, since the whole Linux+HDR thing seems to be pretty new.
Unfortunately the Deck is my only device that can render HDR content, so it'll take somebody else to test if the colors and contrast actually look "correct".

streetpea:
I could at least compare the images by eye. I have an hdr tv i can use and compare screenshots in-game vs streaming.

jbaiter:
If I understood the gamescope code correctly, the HDR support is not limited to the OLED deck's internal screen, but should extend to displays attached via USB-C as well, as long as they declare HDR support in their EDID. So you should be able to confirm it's working by simply having both your deck and the PS5 connected to the display and then switching between them to see if the colors match.

However, this is probably going to take quite a lot of time, I'm basically starting from almost zero with graphics programming and especially color handling, so if there's anybody with experience in these areas, and especially Vulkan, and would like to collaborate (or take it over completely), get in touch :-)

streetpea:
I personally have very limited graphics rendering experience so agree that libplacebo seems like a good idea. 

Also, regarding testing the image I can set my tv to render different inputs on the same screen via a split screen feature so I can render the steam deck screen using remote play at the same time as the native ps stream for comparison purposes.

realdannys:
Really glad to see some guys are already working on this. Ratchet and Clank in HDR on the Deck would be amazing!

jbaiter:
So I've made some progress and got Chiaki to render the 10bit HEVC stream of the PS55 using libplacebo instead of the current OpenGL renderer.
There's only one problem: It seems that in order for the PS5 to send actual HDR10 signals along with the video packets, it needs to be hooked up  to a HDR-capable display, at least according to this official Sony support article (my emphasis):

> Using Remote Play with HDR video quality
> 
> If your Windows PC **and** PS5 console are connected to a display or TV that supports HDR, you can use Remote Play with HDR video quality.

https://remoteplay.dl.playstation.net/remoteplay/lang/gb/ps5_win.html#:%7E:text=If%20your%20Windows%20PC%20and,HDR%5D%20to%20%5BAutomatic%5D

[...]

PR #98, opened 2023-11-22
VAAPI zero copy decoding

nowrep:
Import the decoded surface directly instead of reading it back to system memory and then uploading to GPU again.
Gives around 50% reduced CPU usage.

### DIFF
--- a/gui/include/avopenglwidget.h
+++ b/gui/include/avopenglwidget.h
@@ -11,6 +11,7 @@
 extern "C"
 {
 #include <libavcodec/avcodec.h>
+#include <libavutil/hwcontext_drm.h>
 }
 
 #define MAX_PANES 3
@@ -46,6 +47,7 @@ struct AVOpenGLFrame
 	ConversionConfig *conversion_config;
 
 	bool Update(AVFrame *frame, ChiakiLog *log);
+	bool UpdateDrmPrime(AVFrame *frame, ChiakiLog *log);
 };
 
 class AVOpenGLWidget: public QOpenGLWidget
--- a/gui/src/avopenglwidget.cpp
+++ b/gui/src/avopenglwidget.cpp
@@ -11,6 +11,9 @@
 #include <QThread>
 #include <QTimer>
 
+#include <EGL/egl.h>
+#include <EGL/eglext.h>
+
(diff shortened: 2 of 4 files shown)


streetpea:
Hi @nowrep thanks for putting this together. This seems like a good idea but @jbaiter is working on switching to Vulkan vs OpenGL via libplacebo for HDR support (see #93) so that will replace the avopenglwidget. If this change can be replicated for that updated code (once its done) then that would be much appreciated. Also, if you have graphics programming experience and would like to help with the HDR update using libplacebo feel free to reach out to @jbaiter in #93 .

nowrep:
> ... switching to Vulkan ... so that will replace the avopenglwidget

Sure, when it's ready it won't make much sense to use VAAPI anymore.

> would like to help with the HDR update using libplacebo

I'm not really sure it's even possible to do with current Qt Vulkan API (there doesn't seem to be getter for VkSwapchainKHR).

jbaiter:
> I'm not really sure it's even possible to do with current Qt Vulkan API (there doesn't seem to be getter for VkSwapchainKHR).

You can create the `VkInstance` in libplacebo and then pass it to Qt via `QWindow::setVkInstance` to make the window use it. Then you can use all of libplacebo's API's to set up the swapchain and everything else. As I mentioned in the other thread, I have rendering working using this approach.

As for zero-copy, I think libplacebo already does this transparently for us: https://code.videolan.org/videolan/libplacebo/-/blob/master/src/include/libplacebo/utils/libav_internal.h?ref_type=heads#L1212-1227

However, as already mentioned in the other issue, I don't see any reason against both of the renderers coexisting, the OpenGL implementation is probably going to be faster, since placebo (at least in the default config) does a lot of stuff in addition to the simple decode -> convert pixfmt -> render loop of the `AVOpenGLWidget`, which comes at a cost of more CPU and GPU utilization.

nowrep:
> Then you can use all of libplacebo's API's to set up the swapchain and everything else

Right, if you completely side-step Qt and use it just to create the window surface, makes sense.

> the OpenGL implementation is probably going to be faster

It probably won't be a big difference, if any (or even Vulkan could be faster).

PR #104, opened 2023-11-29
HDR streaming and rendering via libplacebo

jbaiter:
Implements #93.

This adds a second renderer besides the previous OpenGL renderer.
It uses libplacebo with its Vulkan output to draw the frames to the stream window. libplacebo has quite a few nice features, among others:

- Zero-copy rendering if the decoded frame is already residing in GPU memory
- **Rendering of HDR content**
- Post-processing to reduce compression artifacts (especially debanding is nice!)
- API for rendering overlays on the video frame

The renderer can be configured in the settings and still defaults to OpenGL, since on my laptop Placebo is still dropping frames at 4k rendering resolution on my laptop (Ryzen 7 6850U with Rembrandt graphics), even on the fast preset. This is in contrast to the OpenGL renderer, which keeps up nicely without dropping any frames.

Placebo offers three presets with increasing amounts of post-processing. The highest preset e.g. implements debanding, which almost gets rid of the annoying color banding in the PS5's main screen interface. When the Placebo renderer is enabled, users can take their pick between the three presets.

Remaining TODOs:

- [x] Make Placebo renderer a compile-time option
- [x] Fix `AVCodec internal buffer is full removing frames before pushing` errors and resulting glitches when streaming window is not visible
- [x] Documentation


streetpea:
For testing I created a flatpak repo you can access by downloading this file [io.github.streetpea.Chiaki4deck.txt](https://github.com/streetpea/chiaki4deck/files/13454842/io.github.streetpea.Chiaki4deck.txt) and rename it to **io.github.streetpea.Chiaki4deck-hdr.flatpakref** instead of .txt (I changed it to .txt for Github to allow me to upload it here) and then use the following in the directory where its located:
```
flatpak install --user -y io.github.streetpea.Chiaki4deck-hdr.flatpakref
```
After running this file once to test with:
```
flatpak run io.github.streetpea.Chiaki4deck-hdr
```
replace the config file with a copy of your chiaki4deck config file with:
```
cp ~/.var/app/io.github.streetpea.Chiaki4deck/config/Chiaki/Chiaki.conf ~/.var/app/io.github.streetpea.Chiaki4deck-hdr/config/Chiaki/Chiaki.conf
```
and then open it again and select **Placebeo (Vulkan, needed for hdr)** from the Renderer settings on the bottom of the settings window.

Finally, add it to Steam as a non-steam game. Then launch it in game mode (hdr only works in game mode for now [apparently kde is planning hdr support for 2024])

As updates are made to this build I'll update the repo and testers will be able to use flatpak upgrade io.github.streetpea.Chiaki4deck-hdr

The file to create a similar build on your own is scripts/flatpak/chiaki4deck.yaml you just need to update the sources to point to your updated sources or local sources (for more info on how to do this see https://streetpea.github.io/chiaki4deck/diy/buildit/#create-gpg-key-for-signing-your-builds-and-repositories and don't do this on your Steam Deck)

jbaiter:
> The big limitation is going to be with some peoples TVs. Some, like my LG C9, will only signal to the PS5 that HDR is available if it’s on and connected on that input. Some have power saving options that can be turned off so the handshake is there all the time. This is a limitation of the official remote play apps as well.

I'm still testing various HDMI dongles to see if they allow us to work around this, maybe we get lucky 🤞

arshiatn:
Okay, I tested it out for almost 2 hours now.

As you already mentioned @streetpea, hardware decoding doesn't work. HDR works though. Colors look the same like on my LG C2.

--------------------------------------------
Really important to note for everyone that wants to use their Steam Deck OLED for HDR:

You have to go to calibrate the HDR for your Deck OLED if your TV doesn't output same max 1000 nits like Deck OLED. 

In PS5, open up settings => Screen and video => video output => scroll down to Adjust HDR. (It only works outside Remote Play)

My C2 monitor outputs max 800 nits so my settings are 14-14-0. 14 means 14 steps from beginning. 

For Deck OLED you need to use 16-16-0. 

Extra: What each setting outputs:
14 - 14 - 0 | MAX 853 nits
15 - 15 - 0 | MAX 985 nits
16 - 16 - 0 | MAX 1137 nits

To verify if there was a difference, I turned on HDR Waveform graph. (Go to settings on your Deck, display and scroll completly down. There you can change "none" to HDR Waveform.). And yes it makes a difference max nits went up to 1000-1140 with 16-16-0 from 750-850.
Horizon Forbidden West is a great HDR example if you want to try out HDR. (Ratchet and Clank Rift Aprt too)
--------------------------------------------

Something to note that may not matter: red flowers in home menu of Horzion Forbidden West are more "red" than on Deck OLED but it is probably because I am using a custom setting on my C2.

Summary: I played Ratchet and Clank after testing everything out for 10 minutes (1080p+60fps+10000 video bitrate + HDR) and worked really good. It was "stuttering" a bit  at beginning but Lowering bitrate from 15000 to 10000 helped a bit. I think my Firtzbox 6690 is just doing a prank on me. Anyway, after restarting my router, I can actually get 15000 bitrate without any issues, as long I change governor to performance in powertools (a plugin for Deck) or limit cpu min ghz to 2

Extra: Capping Chiaki to 60hz flattens the Frametime graph, so I really recommend that. It smooths the stream.

Edit 1: I played Horizon Forbidden West for 30 minutes now and holy sht, the HDR in this game is phenomenal. The stream had some hiccups here and there but that is my router probably because I have the same problem in chiaki4deck (without HDR).

Edit 2: I tested out if HDR works if my monitor is off or not. (I have to note that power saving is off on it)

TV on with PS5 on => Deck HDR on
TV off while PS5 is on=> Deck HDR on
TV off and PS5 off but turning PS5 on after 1 minute with TV still off => Deck HDR on

=> it works on LG C2 OLED without needing in being on all the time. Great!!

Thank you guys for implementing HDR soooo fast. I never thought I could see HDR in Chiaki before HDR in Moonlight xd.

nowrep:
First patch fixes building against libplacebo lib outside of standard library path. Second patch fixes the QtWayland crash (still crashes if trying to close the window before stream starts, but this time crashes inside chiaki).

<details>
<summary>0001-Use-pkg_check_modules-for-lipblacebo.patch</summary>

```diff
From 07880663e683bb45815aeb6e1d58bb5747db5448 Mon Sep 17 00:00:00 2001
From: David Rosca <nowrep@gmail.com>
Date: Thu, 30 Nov 2023 13:29:40 +0100
Subject: [PATCH 1/2] Use pkg_check_modules for lipblacebo

---
 gui/CMakeLists.txt | 8 +++-----
 1 file changed, 3 insertions(+), 5 deletions(-)

diff --git a/gui/CMakeLists.txt b/gui/CMakeLists.txt
index ed3dca3..5e054f2 100755
--- a/gui/CMakeLists.txt
+++ b/gui/CMakeLists.txt
@@ -7,7 +7,7 @@ if(APPLE)
 	find_package(Qt5 REQUIRED COMPONENTS MacExtras)
 endif()
 find_package(PkgConfig REQUIRED)
-pkg_search_module(LIBPLACEBO REQUIRED libplacebo)
+pkg_check_modules(LIBPLACEBO REQUIRED libplacebo IMPORTED_TARGET)
 
 if(WIN32)
 	add_definitions(-DWIN32_LEAN_AND_MEAN)
@@ -65,11 +65,9 @@ add_executable(chiaki WIN32
 		include/loginpindialog.h
 		src/loginpindialog.cpp)
 target_include_directories(chiaki PRIVATE include)
-target_include_directories(chiaki PRIVATE ${LIBPLACEBO_INCLUDE_DIRS})
 
 add_library(placebo-libav-impl src/libav_impl.c)
-target_include_directories(placebo-libav-impl PRIVATE ${LIBPLACEBO_INCLUDE_DIRS})
-target_link_libraries(placebo-libav-impl ${LIBPLACEBO_LIBRARIES} avcodec avutil avformat)
+target_link_libraries(placebo-libav-impl PkgConfig::LIBPLACEBO avcodec avutil avformat)
 
 target_link_libraries(chiaki placebo-libav-impl)
 
@@ -81,7 +79,7 @@ endif()
 
 target_link_libraries(chiaki Qt5::Core Qt5::Widgets Qt5::Gui Qt5::Concurrent Qt5::Multimedia Qt5::OpenGL Qt5::Svg)
 target_link_libraries(chiaki SDL2::SDL2)
-target_link_libraries(chiaki ${LIBPLACEBO_LIBRARIES})
+target_link_libraries(chiaki PkgConfig::LIBPLACEBO)
 if(APPLE)
 	target_link_libraries(chiaki Qt5::MacExtras)
 	target_compile_definitions(chiaki PRIVATE CHIAKI_GUI_ENABLE_QT_MACEXTRAS)
-- 
2.43.0

```

</details>

<details>
<summary>0002-Create-Vulkan-surface-ourselves-instead-of-using-QVu.patch</summary>

```diff
From 0a65926da5cde620d2837baea60d8b1ea3fa4d46 Mon Sep 17 00:00:00 2001
From: David Rosca <nowrep@gmail.com>
Date: Thu, 30 Nov 2023 14:22:22 +0100
Subject: [PATCH 2/2] Create Vulkan surface ourselves instead of using
 QVulkanInstance

---
 gui/CMakeLists.txt            |  2 +-
 gui/include/avplacebowidget.h |  8 +++---
 gui/include/streamwindow.h    |  2 --
 gui/src/avplacebowidget.cpp   | 49 +++++++++++++++++++++++++++++------
 gui/src/streamwindow.cpp      | 12 +++------
 5 files changed, 49 insertions(+), 24 deletions(-)

diff --git a/gui/CMakeLists.txt b/gui/CMakeLists.txt
index 5e054f2..1bd6513 100755
--- a/gui/CMakeLists.txt
+++ b/gui/CMakeLists.txt
@@ -77,7 +77,7 @@ if(CHIAKI_ENABLE_CLI)
 	target_link_libraries(chiaki chiaki-cli-lib)
 endif()
 
-target_link_libraries(chiaki Qt5::Core Qt5::Widgets Qt5::Gui Qt5::Concurrent Qt5::Multimedia Qt5::OpenGL Qt5::Svg)
+target_link_libraries(chiaki Qt5::Core Qt5::Widgets Qt5::Gui Qt5::GuiPrivate Qt5::Concurrent Qt5::Multimedia Qt5::OpenGL Qt5::Svg)
 target_link_libraries(chiaki SDL2::SDL2)
 target_link_libraries(chiaki PkgConfig::LIBPLACEBO)
 if(APPLE)
diff --git a/gui/include/avplacebowidget.h b/gui/include/avplacebowidget.h
index 548ad7b..8b95aa0 100644
--- a/gui/include/avplacebowidget.h
+++ b/gui/include/avplacebowidget.h
@@ -10,7 +10,6 @@
 
 #include <QWidget>
 #include <QMutex>
-#include <QVulkanInstance>
 #include <QWindow>
 #include <QElapsedTimer>
 #include <libplacebo/options.h>
@@ -36,8 +35,8 @@ class AVPlaceboWidget : public QWindow, public IAVWidget
         AVPlaceboFrameUploader *frame_uploader;
         QMutex frames_mutex;
         QThread *frame_uploader_thread;
+        VkSurfaceKHR surface = VK_NULL_HANDLE;
 
-        QVulkanInstance *vulkan_instance;
         pl_render_params render_params;
         pl_log placebo_log;
         pl_vk_inst placebo_vk_inst;
@@ -48,7 +47,7 @@ class AVPlaceboWidget : public QWindow, public IAVWidget
 
     public:
         explicit AVPlaceboWidget(
-            StreamSession *session, pl_log placebo_log, pl_vk_inst placebo_vk_inst,  QVulkanInstance *qvkinst,
+            StreamSession *session, pl_log placebo_log, pl_vk_inst placebo_vk_inst,
             ResolutionMode resolution_mode = Normal, PlaceboPreset preset = PlaceboPreset::Default
         );
         ~AVPlaceboWidget() override;
@@ -59,6 +58,7 @@ class AVPlaceboWidget : public QWindow, public IAVWidget
         void showEvent(QShowEvent *event) override;
         void resizeEvent(QResizeEvent *event) override;
         static void PlaceboLog(void *user, pl_log_level level, const char *msg);
+        VkSurfaceKHR vkSurface() const { return surface; }
 
     protected:
         ResolutionMode resolution_mode;
@@ -68,4 +68,4 @@ class AVPlaceboWidget : public QWindow, public IAVWidget
         void ToggleZoom() override;
 };
 
-#endif // CHIAKI_AVPLACEBOWIDGET_H
\ No newline at end of file
+#endif // CHIAKI_AVPLACEBOWIDGET_H
diff --git a/gui/include/streamwindow.h b/gui/include/streamwindow.h
index c7dd8bc..f2c6dfe 100644
--- a/gui/include/streamwindow.h
+++ b/gui/include/streamwindow.h
@@ -11,7 +11,6 @@
 #include "avwidget.h"
 
 class QLabel;
-class QVulkanInstance;
 
 class StreamWindow: public QMainWindow
 {
@@ -34,7 +33,6 @@ class StreamWindow: public QMainWindow
 		pl_log placebo_log;
 		pl_vk_inst placebo_vk_inst;
 		pl_vulkan placebo_vulkan;
-		QVulkanInstance *vulkan_instance;
 		QWidget *placebo_widget;
 
 		void Init();
diff --git a/gui/src/avplacebowidget.cpp b/gui/src/avplacebowidget.cpp
index fe4e824..e5b230f 100644
--- a/gui/src/avplacebowidget.cpp
+++ b/gui/src/avplacebowidget.cpp
@@ -13,13 +13,18 @@
 #define PL_LIBAV_IMPLEMENTATION 0
 #include <libplacebo/utils/libav.h>
 
+#include <xcb/xcb.h>
+#include <vulkan/vulkan.h>
+#include <vulkan/vulkan_xcb.h>
+#include <vulkan/vulkan_wayland.h>
+#include <qpa/qplatformnativeinterface.h>
+
 AVPlaceboWidget::AVPlaceboWidget(
-    StreamSession *session, pl_log placebo_log, pl_vk_inst placebo_vk_inst, QVulkanInstance *qvkinst,
+    StreamSession *session, pl_log placebo_log, pl_vk_inst placebo_vk_inst,
     ResolutionMode resolution_mode, PlaceboPreset preset)
     : placebo_log(placebo_log), placebo_vk_inst(placebo_vk_inst), session(session), resolution_mode(resolution_mode)
 {
     setSurfaceType(QWindow::VulkanSurface);
-    this->setVulkanInstance(qvkinst);
 
     if (preset == PlaceboPreset::Default)
     {
@@ -53,8 +58,40 @@ void AVPlaceboWidget::showEvent(QShowEvent *event) {
     QWindow::showEvent(event);
     this->requestActivate();
 
+    VkResult err;
+
+    if (QGuiApplication::platformName().startsWith("wayland")) {
+        PFN_vkCreateWaylandSurfaceKHR createSurface = reinterpret_cast<PFN_vkCreateWaylandSurfaceKHR>(
+                placebo_vk_inst->get_proc_addr(placebo_vk_inst->instance, "vkCreateWaylandSurfaceKHR"));
+
+        VkWaylandSurfaceCreateInfoKHR surfaceInfo;
+        memset(&surfaceInfo, 0, sizeof(surfaceInfo));
+        surfaceInfo.sType = VK_STRUCTURE_TYPE_WAYLAND_SURFACE_CREATE_INFO_KHR;
+        QPlatformNativeInterface* pni = QGuiApplication::platformNativeInterface();
+        surfaceInfo.display = static_cast<struct wl_display*>(pni->nativeResourceForWindow("display", this));
+        surfaceInfo.surface = static_cast<struct wl_surface*>(pni->nativeResourceForWindow("surface", this));
+        err = createSurface(placebo_vk_inst->instance, &surfaceInfo, nullptr, &surface);
+    } else if (QGuiApplication::platformName().startsWith("xcb")) {
+        PFN_vkCreateXcbSurfaceKHR createSurface = reinterpret_cast<PFN_vkCreateXcbSurfaceKHR>(
+                placebo_vk_inst->get_proc_addr(placebo_vk_inst->instance, "vkCreateXcbSurfaceKHR"));
+
+        VkXcbSurfaceCreateInfoKHR surfaceInfo;
+        memset(&surfaceInfo, 0, sizeof(surfaceInfo));
+        surfaceInfo.sType = VK_STRUCTURE_TYPE_XCB_SURFACE_CREATE_INFO_KHR;
+        QPlatformNativeInterface* pni = QGuiApplication::platformNativeInterface();
+        surfaceInfo.connection = static_cast<xcb_connection_t*>(pni->nativeResourceForWindow("connection", this));
+        surfaceInfo.window = static_cast<xcb_window_t>(winId());
+        err = createSurface(placebo_vk_inst->instance, &surfaceInfo, nullptr, &surface);
+    } else {
+        Q_UNREACHABLE();
+    }
+
+    if (err != VK_SUCCESS) {
+        qFatal("Failed to create VkSurfaceKHR");
+    }
+
     struct pl_vulkan_swapchain_params swapchain_params = {
-        .surface = vulkan_instance->surfaceForWindow(this),
+        .surface = surface,
         .present_mode = VK_PRESENT_MODE_FIFO_KHR,
     };
     placebo_swapchain = pl_vulkan_create_swapchain(placebo_vulkan, &swapchain_params);
@@ -76,10 +113,6 @@ void AVPlaceboWidget::showEvent(QShowEvent *event) {
 }
 
 void AVPlaceboWidget::Stop() {
-    // Need to stop the frame uploader thread before the widget is destroyed,
-    // otherwise it segfaults inside the QtWaylandClient
-    // FIXME: Sometimes it still crashes on exit deep inside of Qt, why is this?
-    //        Test on Steam Deck/gamescope if it happens there as well...
     if (frame_uploader_thread) {
         frame_uploader_thread->quit();
         frame_uploader_thread->wait();
@@ -201,4 +234,4 @@ void AVPlaceboWidget::PlaceboLog(void *user, pl_log_level level, const char *msg
     }
 
     chiaki_log(log, chiaki_level, "[libplacebo] %s", msg);
-}
\ No newline at end of file
+}
diff --git a/gui/src/streamwindow.cpp b/gui/src/streamwindow.cpp
index 8d7146d..08e5b9c 100644
--- a/gui/src/streamwindow.cpp
+++ b/gui/src/streamwindow.cpp
@@ -45,8 +45,6 @@ StreamWindow::~StreamWindow()
 	pl_vulkan_destroy(&placebo_vulkan);
 	pl_vk_inst_destroy(&placebo_vk_inst);
 	pl_log_destroy(&placebo_log);
-	vulkan_instance->destroy();
-	delete vulkan_instance;
 }
 
 void StreamWindow::Init()
@@ -107,15 +105,11 @@ void StreamWindow::Init()
 			};
 			placebo_vk_inst = pl_vk_inst_create(placebo_log, &vk_inst_params);
 
-			vulkan_instance = new QVulkanInstance();
-			vulkan_instance->setVkInstance(placebo_vk_inst->instance);
-			vulkan_instance->create();
-
 			auto widget = new AVPlaceboWidget(
-				session, placebo_log, placebo_vk_inst, vulkan_instance,
+				session, placebo_log, placebo_vk_inst,
 				resolution_mode, connect_info.settings->GetPlaceboPreset());
 			widget->installEventFilter(this);
-			VkSurfaceKHR surface = vulkan_instance->surfaceForWindow(widget);
+			VkSurfaceKHR surface = widget->vkSurface();
 
 			struct pl_vulkan_params vulkan_params = {
 				.instance = placebo_vk_inst->instance,
@@ -413,4 +407,4 @@ void StreamWindow::PlaceboLog(void *user, pl_log_level level, const char *msg)
     }
 
     chiaki_log(log, chiaki_level, "[libplacebo] %s", msg);
-}
\ No newline at end of file
+}
-- 
2.43.0

```

</details>

streetpea:
@nowrep is this a fix for
```
CreateSwapchainKHR: Creating swapchain for non-Gamescope swapchain.
Hooking has failed somewhere!
You may have a bad Vulkan layer interfering.
Press OK to try to power through this error, or Cancel to stop
```
or something else?

nowrep:
No, that wouldn't fix this.

streetpea:
@jbaiter how do I tell if the ps5 is outputting hdr? I can tell if the Steam Deck is outputting hdr via the badge and the turning on the hdr view in display settings but how do I know if the PS5 is outputting hdr or if libplacebo is outputting hdr from sdr content? I ask this because I noticed that even when I unplug my hdmi cable from my tv hdr is still outputted on the Steam deck when h265 (hdr) is selected

jbaiter:
Wow, thanks a ton @nowrep!! 🙏🏾 🙏🏾  I merged your changes into the PR!

> @jbaiter how do I tell if the ps5 is outputting hdr?

Two ways to check as far as I could find out:
1. Enter Settings, navigate to "Screen and Video" entry, start controller attached to PS5 (thereby terminating the streaming session), hit X a few times, wait 20-30s, restart streaming session, try to adjust HDR settings.
2. Start a game that allows configuring HDR in their settings (Cyberpunk 2077 e.g. has a menu for that), it will only be available when HDR is enabled and active.

As for my experiments with HDMI dongles, no luck so far. I've now tried a headless dongle (-> 👎🏾) and a dongle with passthrough (-> 👎🏾 as well). I'm hesitant to purchase more of those things, since they all seem to only emulate the EDID, wich doesn't seem to be enough.

streetpea:
@jbaiter I was able to use 2) with the HDR brightness setting in Ghost of Tsushima and hdr is enabled for lg oled c7 whenever the tv is on (doesn't need to be on the ps5 tv input). This works for me since I usually remote play while someone else is watching TV or I'm watching TV and playing simultaneously. I also noticed that after activating it still works for quite some time (even through setting the PS to sleep mode and unplugging the hdmi adapter) until I manually switch H265 without hdr and then I had to turn the tv back on to re-enable hdr with the ps5. However, I'm not sure how long it will remain enabled after initially setting it. I could test this by setting the tv on and then unplugging the hdmi cable from the PS5 and testing if it was still enabled after 1 hour, 2 hours, etc. though I'm not sure if that would be TV specific. (I'm thinking perhaps it's possible to enable via a turned on TV and then unplug the HDMI cable and play for a while without needing the TV to be on).

jbaiter:
I noticed that error whenever I was hiding the stream window in my wayland session, e.g. by switching to another virtual screen. It seems that this makes the `RenderFrame` function take too long, which leads to a full buffer on the libavcodec side (which would be emptied by the calls to `chiaki_ffmpeg_decoder_full_frame` in the frame uploader otherwise).

imacomputa:
> @jbaiter I was able to use 2) with the HDR brightness setting in Ghost of Tsushima and hdr is enabled for lg oled c7 whenever the tv is on (doesn't need to be on the ps5 tv input). This works for me since I usually remote play while someone else is watching TV or I'm watching TV and playing simultaneously. I also noticed that after activating it still works for quite some time (even through setting the PS to sleep mode and unplugging the hdmi adapter) until I manually switch H265 without hdr and then I had to turn the tv back on to re-enable hdr with the ps5. However, I'm not sure how long it will remain enabled after initially setting it. I could test this by setting the tv on and then unplugging the hdmi cable from the PS5 and testing if it was still enabled after 1 hour, 2 hours, etc. though I'm not sure if that would be TV specific. (I'm thinking perhaps it's possible to enable via a turned on TV and then unplug the HDMI cable and play for a while without needing the TV to be on).
> 
> I also ran into a bug where the screen stopped updating but sound and controls still worked (I could hear the sound of navigating the menus despite not being able to see them). The only thing I could see in the logs was:
> 
> ```
> AVCodec internal buffer is full removing frames before pushing
> ```
> 
> The connection was otherwise stable as there were no dropped frames being reported. Here is the log from the run: [chiaki_session_2023-11-30_14-40-42-733733.log](https://github.com/streetpea/chiaki4deck/files/13518655/chiaki_session_2023-11-30_14-40-42-733733.log)
> 
> I also noticed that when the prompt to ask if you want to quit the session (if you have it set to Ask when you disconnect in the settings) comes up this floods the logs and this also happens at the beginning of the stream session. Not sure if those are related though. People have reported an issue with the stream stopping but those logs look very different with the controller actually not responding and connections getting resent while in this case the stream connection seems fine (though I can enable debug logs and try to replicate this to make for sure) and it seems to be something with displaying the frames via the avplacebowidget. Now that I have verbose logging on I'll post an updated log if I can replicate this.

I believe HDR works with the TV off in games that use the system level HDR settings.  I played Spider-Man 2 and Ratchet without the TV on (LG C9) and HDR was definitely working. Neon signs in Spider-Man burned my retinas. 

Then I tried Horizon FW and it was not working until I turned the TV on.

jbaiter:
No success in patching the HDMI capture stick's EDID, so this will probably have to be documented as "supported only when an HDR display is attached to the PS5" for now.

Other than that, I've been playing Cyberpunk and Returnal both with HDR enabled for an hour without problems on my OLED deck with VAAPI hardware decoding and the "High" preset, looks great and the GPU was never using more than 3W, CPU even less 👍🏾 HDR debugging with the waveform and gamut graphs showed that it was properly working and reaching 1000 nits for bright spots throughout 🥳 😎 🌞 💡 

I've updated the PR with a few TODOs that should be done before it leaves the Draft state, can you think of any more @streetpea @nowrep?

streetpea:
@jbaiter I would probably want to add a build flag for the placebo stuff so it can be built without it for the appimage build (right now the appimage build isn't playing nice with libplacebo vulkan build) and windows build.

jbaiter:
I "fixed" the dropped frames error by changing the log level to verbose (can be helpful for debugging issues) and added an API for retrieving total/dropped counts from the ffmpeg decoder (could be used for an overlay or a session summary).

The reason for it happening seems to be that `pl_swapchain_swap_buffers` blocks in case the window is not rendered by Wayland at the moment. Glitches probably occur when I-Frames were missed during that time?

The Placebo renderer can now also  be disabled at compile-time by adding `-DCHIAKI_ENABLE_PLACEBO=OFF` when generating the makefiles.

nowrep:
> I "fixed" the dropped frames error by changing the log level to verbose (can be helpful for debugging issues) and added an API for retrieving total/dropped counts from the ffmpeg decoder (could be used for an overlay or a session summary).

I fixed that by not stalling the decode thread here: https://github.com/jbaiter/chiaki4deck/pull/4

Also fixed the gamescope wsi error when running in desktop mode, and it now shows the chiaki logo when waiting for rendering first frame instead of showing black screen.

With that, I think it's good to go. I would also say that the OpenGL renderer should be completely disabled when building with placebo, because it's all around worse (or at the very least, not make it default).

jbaiter:
One additional thing I observed: The tone mapping from HDR to SDR seems to work now as well, when I turn HDR on my non-HDR laptop, the colors no longer look washed out but seem fine!

jbaiter:
Brillaint @nowrep 🥳 

> it now shows the chiaki logo when waiting for rendering first frame instead of showing black screen.

Lovely, much better UX! Do you think you can resize it to the viewport size? It's currently cropped if the window size doesn't correspond to the expected stream resolution.

nowrep:
> Do you think you can resize it to the viewport size? It's currently cropped if the window size doesn't correspond to the expected stream resolution.

Yeah it doesn't really work great on desktop because the resizing doesn't work at all. This also applies for the video stream, ideally we should hold on the frame until receiving new one so we can redraw on window resize. But it's not so noticeable with video stream because it will re-render at correct size when next frame arrives.

jbaiter:
> With that, I think it's good to go. I would also say that the OpenGL renderer should be completely disabled when building with placebo, because it's all around worse (or at the very least, not make it default).

Do you agree @streetpea? I think I do, given that even the "high quality" preset is absolutely fine regarding energy usage, and the image quality is improved (especially with the debanding).

@nowrep One issue I've encountered after the move of the rendering to the main thread is that the main window becomes unresponsive after closing the streaming session.

jbaiter:
> Also, do you think the high quality preset should be the default instead of normal given the debanding since it still doesn't seem like it's causing that big of a battery hit given the advantage of the debanding.

👍🏾 I do, I think getting rid of the annoying banding is worth the at most a few hundred milliwatts in increased energy usage 🙃

streetpea:
@chipgw mentioned he no longer sees hdr activating in #93 and I also see it. I just deployed this version (where hdr doesn't work anymore) using the latest changes up to now from the jbaiter repo as well as the latest chiaki4deck.yaml. This probably has to do with switching from using a global environment variable to activate the vulkan layer to the qt environment variable.

panthurrr:
I updated my flatpak after reading chipgw's comment and I still have HDR.  So if your deployment was very very recent maybe I'm still one behind.

Also, I would recommend a warning or disabling configurations that are likely to enter Chiaki's FEC; my stream will flash about a third of the screen white at full 1000 nits which should be avoided when it is not part of the actual frame data.

HDR+1080p+vaapi works amazing while HDR+720p+none/vulkan is at least stable.  Dropping the brightness of any error corrected frames seems fine too though vaapi in HDR hasn't had any issues for me yet.

streetpea:
@panthurrr are you using the flatpak repo I made or building it yourself? Here is a log with logging as well as output to the terminal logged and you can see the Gamescope WSI is loading and supposedly enabling HDR so I don't know why the other surfaces aren't appearing. Checking my vulkaninfo flatpak it sill shows the 12 surfaces with the vulkan layers in gamemode so not sure what's going on with the chiaki4deck flatpak. Here is the log
[HDRNotActivating.txt](https://github.com/streetpea/chiaki4deck/files/13533634/ChiakiLOGDOG.txt)

gui/src/avplacebowidget.cpp
```diff
@@ -44,6 +44,9 @@ AVPlaceboWidget::AVPlaceboWidget(
         CHIAKI_LOGI(session->GetChiakiLog(), "Using placebo high quality preset");
         render_params = pl_render_high_quality_params;
     }
+
+    QImageReader logoReader(":icons/chiaki.svg");
```
nowrep, line 48: It's SVG image so that's why QImageReader was used to always render the SVG at requested size. So now it will be blurry due to scaling in RenderPlaceholderIcon.

nowrep:
You don't get HDR because of this:

```
[I] [libplacebo] Available surface configurations:
[I] [libplacebo]     0: VK_FORMAT_B8G8R8A8_SRGB                  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[I] [libplacebo]     1: VK_FORMAT_B8G8R8A8_UNORM                 VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
[I] [libplacebo] Picked surface configuration 1: VK_FORMAT_B8G8R8A8_UNORM + VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
```

not sure why it happens (there should be more formats, notably VK_FORMAT_R16G16B16A16_UNORM and VK_FORMAT_A2B10G10R10_UNORM_PACK32).

gui/src/avplacebowidget.cpp
jbaiter: I tested it on my display and it didn't seem that blurry to me, but I'll revert it. It's not like we're redrawing that logo every frame anyway.

nowrep:
> @nowrep One issue I've encountered after the move of the rendering to the main thread is that the main window becomes unresponsive after closing the streaming session.

@jbaiter Is this also on desktop or only on deck in gamescope? If gamescope then I haven't figured out yet how to cleanly stop stream other than force quit app :smiling_face_with_tear:

gui/src/main.cpp
```diff
@@ -61,6 +61,15 @@ int real_main(int argc, char *argv[])
 	QApplication::setApplicationDisplayName("chiaki4deck");
 	QApplication::setDesktopFileName("chiaki4deck");
 
+	if(qEnvironmentVariable("XDG_CURRENT_DESKTOP") == "gamescope")
+	{
+		QString layers = qEnvironmentVariable("VK_INSTANCE_LAYERS");
+		if(!layers.isEmpty())
+			layers.append(",");
```
nowrep, line 68: Separator here needs to be `:`, not `,`.

panthurrr:
@streetpea using the build from your .flatpakref and upgrading it, I'll get a local build going too and see if that works but here's the log with vaapi+HDR+1080p works great to me.  I tried upgrading it again just now too and still works with HDR.

[HDR with vaapi](https://raw.githubusercontent.com/panthurrr/chiaki4deck/logs/chiaki_session_2023-12-01_21-33-35-197197.log)

jbaiter:
> @jbaiter Is this also on desktop or only on deck in gamescope? If gamescope then I haven't figured out yet how to cleanly stop stream other than force quit app :smiling_face_with_tear: 

It was on my laptop, yeah 😅

nowrep:
Can't reproduce on deck or on my desktop (sway). I assume the main thread is blocked, can you check where in gdb?

streetpea:
@nowrep you can close the stream cleanly from game mode on deck using ctrl+q. This is mapped to one of the back button in the chiaki4deck+mic layout. If you aren't using the automation the settings menu will pop up after closing. If you are using the automation it will just exit out of the app.

nowrep:
@streetpea Yeah, I've loaded the controller layout and it works fine.

streetpea:
An update on hdr not activating. I noticed hdr did activate 1 time when I was gathering logs so I thought maybe it was something that changed in my environment rather than the code. I switched to the commit made a few days ago (which I had tested numerous times and confirmed hdr working) and it still doesn't work so I don't think it's related to code changes.

nowrep:
@panthurrr Nothing wrong in the log, as long as you selected H265 HDR codec (probably good idea to add this to log).

>  Dropping the brightness of any error corrected frames seems fine too

This would probably be a good idea to look into.

Hasney69:
I was getting bad audio crackling today, both on this branch and the main branch, so I was diagnosing the problem. Went into the Steam Deck settings while the HDR stream was up and when I came back, my screen was basically green. Seems to be reproducible as a minor bug.

Step 1) Press Steam button and go to settings
2) Go to internet
3) Press A on your WiFi connection for more information
4) Press B to go back and then go back to your streaming session

Screenshot: https://media.discordapp.net/attachments/853001265013719042/1180443067237548052/20231202_093824.jpg?ex=657d7046&is=656afb46&hm=cbbc1816bfa42d8d36bdc6ba0cb0f7246d413863344085fd78020217f6f239ec&

arshiatn:
I updated about 1 hour ago and the differences to last build I updated to (10-11 hours ago):

You guys added Chiaki Logo and eveerything works like last night. I just played Demon Souls for 30 minutes with following settings and I had no problem. (Some "fps drop" every minute or so but dropping bitrate from 20000 to 11000 fixed it => my wifi on Deck got 2 out 4 bars. )

PS5 1080p HDR 60fps --- 11000 video bitrate --- Vulkan

No problem as I have said before. HDR still works and gets detected. It looks just like last night.


Furthermore I don't know how @panthurrr is using vaapi+HDR+1080p but vaapi or any other hardware recording (two of them crash Chiaki) gives me green screen, even when I lower bitrate to 5000 and go near my router. vaapi + non HDR + opengl doesn't work either. On other hand Vulkan works great compared to "none" HardwareEncoding.


---
To "white Flashes". I got those last night when I was trying 30000-40000 video bitrate and it was literally blinding me with 1000nits flashes every 5s for couple of minutes (I wanted to save the game in Ghost of Tsushima so I endured it). Reducing the bitrate back to 15000 last night, fixed this "white flashing" issues @panthurrr described. ==> lowering bitrate could help and maybe your connection is bad like me xd. (PS5 1Gbit LAN connection to Router and Router is giving me 0.5-1.4gbit 5ghz connection to my Deck through 40-45cm of a concrete wall)

With my 11000  video bitrate setting, I don't have any "white flashes" or "green" screen issues. However I saw similar green screen thingy like @Hasney69 when I went to Settings --> Display --> scroll down --> none to HDR Waveform (to check what max nits is) ---> and back to the game. However it only happend once after doing it like 15 times or so.

nowrep:
> On other hand Vulkan works better compared to "none" HardwareEncoding for HDR.

Anything other than `vaapi` doesn't work and will fallback to software decoding (aka `none`).

arshiatn:
> > On other hand Vulkan works better compared to "none" HardwareEncoding for HDR.
> 
> Anything other than `vaapi` doesn't work and will fallback to software decoding (aka `none`).
> 
> The green video issues with hardware decoding is simply because hardware decoding is pretty bad at decoding corrupted streams - software decoder will use gray color for invalid frames / parts of frame but on hw decode you get green.

Wait so my connection was still bad when I was sitting infront of my router and that is why I still was getting green screen, right? If I somehow fix this connection thingy issue, I could be able to use vaapi, right?

Weirdly enough, I saw my deck use 8-9w (vulkan) instead of 7-8w (none) so I thought that it was working. Alright, I got you.

scripts/flatpak/chiaki4deck.yaml
```diff
@@ -198,5 +199,5 @@ modules:
         branch: chiaki4deck
     post-install:
       - "install -Dm755 ../scripts/psn-account-id.py /app/bin/psn-account-id"
-      - "install -Dm755 ../scripts/vulkan-gamescope-layers/VkLayer_FROG_gamescope_wsi.x86.json /app/vulkan-layers/VkLayer_FROG_gamescope_wsi.x86_64.json"
+      - "install -Dm755 ../scripts/vulkan-gamescope-layers/VkLayer_FROG_gamescope_wsi.x86.json /app/vulkan-layers/VkLayer_FROG_gamescope_wsi.x86.json"
```
nowrep, line 202: Only `x86_64` layer is used.

streetpea:
I was able to figure out what was causing hdr mode not to activate for me. I recently switched to using the Chiaki-launcher.sh script to launch chiaki4deck-hdr. It turns out when launching chiaki via the cli like
```flatpak run io.github.streetpea.Chiaki4deck-hdr --passcode "4400" --stretch stream PS5-411 192.168.1.16```
(non-steam game set to flatpak command set to `run io.github.streetpea.Chiaki4deck-hdr --passcode "4400" --stretch stream PS5-411 192.168.1.16`) it doesn't activate hdr despite saying gamescope is adding hdr over stdout. Weirdly, it still activates hdr if you are connected to an external monitor when launching. This means currently if you use the Chiaki-launcher.sh script or just launch via the cli option you won't get hdr activating for you unless pluggMayeed into an external monitor. Not exactly sure what's causing this. Maybe @nowrep or @jbaiter have a better idea.

nowrep:
It's because the gamescope layer thinks it can't bypass xwayland in that case. Looking into it.

nowrep:
I can't reproduce, but it may have something to do with window sizes mismatch (AVPlaceboWidget is actually a "subsurface").

Try this hacky patch: https://github.com/nowrep/chiaki4deck/commit/fbeea2a8e75d52755425b0f89d844fece3db66e6

streetpea:
@nowrep you should be able to reproduce if you try launching with the stream command as your non-steam game instead or downloading the [gen-launcher.sh script](https://raw.githubusercontent.com/streetpea/chiaki4deck/main/scripts/gen-launcher.sh) replacing io.github.streetpea.Chiaki4deck with io.github.streetpea.Chiaki4deck-hdr and then running it and then adding the resulting `~/.var/app/io.github.streetpea.Chiaki4deck-hdr/config/Chiaki/Chiaki-launcher.sh` as a non steam game and launching it.

nowrep:
I did it exactly like that (so it opened only the stream window) and hdr still worked.

streetpea:
@nowrep did you have a monitor attached? Hdr still works when I have a monitor attached but doesnt if I don't when launching via the stream command.

nowrep:
No I don't have monitor attached.

streetpea:
Which branch of steamos are you on?

nowrep:
Stable (3.5.7).

streetpea:
@nowrep it's working now using the stream command when I use your swapchain-resize branch 👍

thenayr:
FWIW I've never used Chiaki before, installed this last night and ran flawlessly with full HDR on every game I tested.  PS5 is hardwired LAN connection, was able to push my bitrate as high as 80k with no stuttering and beautiful (practically native looking) picture.  Thanks!

jbaiter:
Thank you for the documentation @streetpea!
And much thanks to you @nowrep for all the help and fixing my stupid bugs ❤️

I think this is ready for review now. The code could probably take some more cleaning up and some more testing with a final build would be great as well. Especially if the occasional white flashes are tolerable with HDR enabled or not. I had a look around the code base where one might change the color to black, but haven't had much luck. Does anybody know what needs to happen for the white flashes trigger vs green blocking in case of a corrupted/missing frame?

streetpea:
I just updated the testing flatpak to the latest build. You can update to the latest version via the discover store or `flatpak upgrade io.github.streetpea.Chiaki4deck-hdr` from the konsole

streetpea:
Another thing that I've noticed is if you set the `Action on Disconnect` to `Ask` and bring up the menu (using ctrl+q) or hitting x on the screen and then try to go back to the session the screen will be frozen (you can still hear audio from what you're doing in the background) but the logs just print `AVCodec internal buffer is full removing frames before pushing` so it seems like the the libplacebo library isn't able to process the frames for the rest of the session when this happens. This doesn't happen with the opengl renderer.

Originalimoc:
Cool, can't wait to test(High-five)

streetpea:
Regarding the flashing frames could we just drop those frames instead of playing them so it lags during that time versus flashing? Is there a specific condition that I associated with the flashing frames?

panthurrr:
> Regarding the flashing frames could we just drop those frames instead of playing them so it lags during that time versus flashing? Is there a specific condition that is associated with the flashing frames?

You can try dropping the corrupt frames instead of flushing them by comment out this line:

https://github.com/streetpea/chiaki4deck/blob/hdr/lib/src/videoreceiver.c#L116

I've also attempted swapping the frame_index_cur with the frame_index_prev_complete before returning from that if statement but there's still a lot of dropped frames which turn the entire screen grey.  

For now choosing vaapi produces a better user experience to me than software decoding; might be preferred to only allow vaapi with HDR on if that's the only hw decoder that works.

jbaiter:
This is just a (probably stupid) idea, but couldn't we keep around a static black frame that we insert whenever we encounter a corrupt frame? Black should be a lot less irritating than white or gray on the OLED screen.

Originalimoc:
Second this, using VAAPI on Deck gamemode, right now(non hdr ver) video is much choppier than Android PSPlay, same WiFi network, PS5 on wired LAN, can easily support 100Mbps.

streetpea:
@nowrep i tend to agree that better handling streaming errors is outside of the scope of this pr so I created a new issue for it #108. Having said that, there were complaints about discomfort particularly when the flashes occurred during HDR streaming because of the increased brightness so if your branch does fix that it could be good to add your change as a hot fix for that.

nowrep:
> Are they any other changes we should add before merging this pr?

Should be good now, only need to [fix build](https://github.com/nowrep/chiaki4deck/commit/df1df19b4cd518f2362684a690831f5267268ed5)

jbaiter:
Merged your latest change @nowrep! Many thanks for all the work ❤️

PR #107, opened 2023-12-05
Add menu item for creating shortcut shell scripts

Nikorag:
Rather than using the shell script provided in the scripts directory, I've added a context menu item for registered consoles which will create the shell script using templating.

Ideally I'd like to automatically add this to Steam by modifying the shortcuts.vdf but this is a first step.


streetpea:
Thx @Nikorag I'll take a look this weekend.

Nikorag:
Just spotted a few errors in the templates. Missing double quotes before the variables. I'll add them back in and update the PR.

Nikorag:
> Just spotted a few errors in the templates. Missing double quotes before the variables. I'll add them back in and update the PR.

All Done - Not had a chance to test it on an actual deck, but the script output looks good to me.

PR #109, opened 2023-12-09
Add to steam

Nikorag:
This is based off my other Pull Request https://github.com/streetpea/chiaki4deck/pull/107 but removes the added step of creating the shell script and adding to steam and adds it directly to Steam.

It's using the Steam Rom Manager's API key to steam grid DB. probably needs replacing with a fresh one for Chiaki4Deck but it was available in github so i figured it was fair game.


Nikorag:
I'm a Java developer as a profession and haven't touched C since university. Ive NEVER touched c++ so had to teach myself as I went with this. If you've got any comments at all please let me know and I'll gladly re-do any of it or change anything. (I plan to remove the author commits and fix the conflicts when I'm next at my laptop)

Happy to help update the documentation too if that's a pre-req for merging this.

Just excited to be involved! Cheers for considering my pull request.

streetpea:
@Nikorag sounds good I'll look at your pr soon (I ended up running out of time to take a look this weekend). I can also help with updating the docs before merging the PR.

streetpea:
Also do you think we could also add the images to Steam so users don't have to load the artwork themselves upon installation.

Nikorag:
Hey! Don't worry about not getting around to it, life gets in the way and this is a passion project, not your day job.

While I'd say "yes" it does supersede the other PR, I'd value your opinion on if there's benefit to keeping both options (Adding directly to steam AND just saving the script somewhere). I wasn't sure if adding directly to steam was even going to be possible, hence raising the other one first. Working with the binary vdf files was fiddly and I ended up writing the parsing and serializing logic myself as I couldn't find a suitable library.

As for artwork, that is included, the dialog which shows when clicking "Add to Steam" allows you to pick from artwork pulled from SteamGridDB for "PS4 Remote Play". I tried adding Chiaki, PlayStation 4 and PlayStation 5 too but SGDB's API doesn't seem to like returning more than 4 images when searching by multiple game IDs and every page after 0 returns results with empty bodies. I'll raise it with them.

If you didn't get around to merging this weekend, I think I'd like to mark this PR as a draft and have a go at tidying up the UI a little.

As I've said, I'm excited that anything I've done is being considered for such a cool project so please give me your honest feedback about this feature.

Nikorag:
Right. I've updated the UI:

Screenshot 2023-12-11 at 16 50 55

You can also now upload your own custom images for each type of Artwork. As well as select options from SteamGridDB from the dropdown menu.

Nikorag:
Hi @streetpea - Thanks for your feedback. I've addressed some of your points:

1. If Steam isn't found it now gives the option to save it elsewhere.
2. Using CHIAKI_LOGE and CHIAKI_LOGI - my VDFParser as an entirely static namespace was tricky as I need to pass the pointer to the log around but it works. Perhaps if it gets too fiddly we can move it to be a class we instantiate but for now it works and it's not too inelegant.
3. Haven't done this one. I'm not sure I completely understand the requirement, nor can I even find the controller_base directory on my mac. Are you able to assist at all? It feels like a value-add on top of this so hopefully it won't block the feature.
4. I've created a common branch for a new class JSONUtil which performs all of the GET/POST logic and pulling the json attributes. I'll apply this to my other PR too. I've also replaced the download file logic with QT so I can remove Curl entirely and the CMAKE property too.

streetpea:
This should be fine. See 2.1 which applies to this here: https://softwarefreedom.org/resources/2007/gpl-non-gpl-collaboration.html

streetpea:
I can definitely build it for the Steam Deck. I'll do that tonight and post instructions for you to download. In the meantime you could try following these instructions to build it. https://streetpea.github.io/chiaki4deck/diy/steamdeckdev/

streetpea:
I tested it on Steam and it works nicely.

You can also test it if you want.

Download this file [io.github.streetpea.Chiaki4deck.txt](https://github.com/streetpea/chiaki4deck/files/13454842/io.github.streetpea.Chiaki4deck.txt) and rename it to **io.github.streetpea.Chiaki4deck-hdr.flatpakref** instead of .txt (I changed it to .txt for Github to allow me to upload it here) and then use the following in the directory where its located:
```
flatpak install --user -y io.github.streetpea.Chiaki4deck-hdr.flatpakref
```
After running this file once to test with:
```
flatpak run io.github.streetpea.Chiaki4deck-hdr
```
replace the config file with a copy of your chiaki4deck config file with:
```
cp ~/.var/app/io.github.streetpea.Chiaki4deck/config/Chiaki/Chiaki.conf ~/.var/app/io.github.streetpea.Chiaki4deck-hdr/config/Chiaki/Chiaki.conf
```

Nikorag:
Hi @streetpea, thanks for this! Glad we're moving closer to getting it merged. I'll look at the best way to guide the player towards adding to steam on registration when I get time. The only three things I had thought of that hadn't been covered yet (perhaps you could help with) is:

- We should probably put a disclaimer about the artwork, that it is taken from SGDB and we're not responsible for its content
- I wasn't able to test what happens when the steam game's name has special characters in it in regards to the controller config vdf. I know to lowercase the game name but I'm unclear on what to do with things like apostrophes in "Jamie's PS4" for example.
- I started on some logic to make the shell script compatible with any *nix system by making "executable" one of the items in the map but I cannot compile chiaki4deck with the cli on my
Mac due to missing libraries. I will probably look to do this at some point and add in the option of generating a .bat file for windows although this could easily be a second PR once this one is merged. It feels like value add more than a requirement.

I'll let you know once I've done the popup and I look forward to hearing back from you. As always, excited to be helping out!!

Nikorag:
Hi @streetpea 

I've added your suggestion of asking if you want to add to steam after registering. I've also amended the scripts to get the executable dynamically (need to test this on linux if you wouldn't mind doing the honours).

I looked at the controller config and it looks as though you just escape the characters which I believe the vdf_parser library will do for us:

```
        "harry potter and the sorcerer\'s stone"
        {
                "autosave"              "1"
        }
```

I've also added a small line of text in the window which explains where the images are coming from.

Only thing I've thought of which is a small improvement is to stop the requests to SteamGridDB blocking the dialog rendering but moving to this multi-threaded approach requires a bit more thought. I think i'm happy with it as-is and to add this as a later improvement.

Nikorag:
I've also now added the asynchronous loading of images after the dialog loads.

streetpea:
I made a pr to your repo for the case where the console nickname bas special characters

Nikorag:
Merged!

streetpea:
I've been testing this out and it seems good now. The only thing I noticed was that you create the script here (https://github.com/Nikorag/chiaki4deck/blob/60ddae277b777ca6ac35d6030cb68daa60acfb1c/gui/src/shortcutdialog.cpp#L425/) if you have Steam but have logic here for other installs (besides flatpak) https://github.com/Nikorag/chiaki4deck/blob/60ddae277b777ca6ac35d6030cb68daa60acfb1c/gui/src/shortcutdialog.cpp#L339 so perhaps you should also check if that flatpak directory exists and if it doesn't ask the user if they want to save somewhere else as well?

Nikorag:
That's done for you

streetpea:
thanks for changing that .. just made a pr to fix the includes on one file you can either merge or manually add it yourself since it's just a few lines. Other than that, I think if possible not using the event loop to block requests would be nice (same as the other pr). Besides that this lgtm now

Nikorag:
I moved to the async fetching of images (which is much faster) as well as abstracting out the ugly duplicated blocks for each artwork type into it's own re-usable class.

Looks a lot tidier now.

Could be tidier still, but for now I'm happy.

streetpea:
When testing this I noticed that if the network gets disconnected when adding the steam shortcut it results in a segfault. To replicate you can click add to Steam and then disconnect your network connection.

Nikorag:
That's probably due to it failing to download the images for the artwork. Should be a simple one to fix! Will take a look at the same time as tidying up the rest of the code.

nowrep:
These scripts are very poor UX, you'll be sitting at black screen with no indication what's going on. Also what's the issue with just selecting console to connect to from UI (and maybe waking it up from there)? Anyway, this should definitely be integrated into UI to display status and be able to cancel the connecting process.

Adding a Steam entry is good thing, but maybe it can be simplified to just set our artwork instead of letting user to choose.

scripts/flatpak/chiaki4deck.yaml
```diff
@@ -204,6 +204,20 @@ modules:
       url: https://gitlab.xiph.org/xiph/speexdsp.git
       tag: SpeexDSP-1.2.1
 
+  - name: wireless-tools
```
nowrep, line 207: You can just use `dbus-send` to communicate with NetworkManager instead of pulling yet another dependency.

gui/src/shortcutdialog.cpp
```diff
@@ -0,0 +1,311 @@
...
+    if (_NSGetExecutablePath(buffer, &size) == -1) {
+        // Buffer is not large enough, allocate a larger one
+        char* largerBuffer = new char[size];
+        if (_NSGetExecutablePath(largerBuffer, &size) == 0) {
+            return largerBuffer;
+        }
+    } else {
+        return buffer;
+    }
+#elif defined(__linux)
```
nowrep, line 88: This part is all wrong. First it assumes the flatpak name and the other path won't work for appimage either I think (it would return the executable path inside the temporary appimage mount).

gui/src/shortcutdialog.cpp
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    fileText = fileText+compileTemplate("launch.tmpl", paramMap);
+
+    bool steamExists = SteamShortcutParser::steamExists();
+    std::string executable = getExecutable();
+
+    std::string filePath;
+    if (steamExists && executable.rfind("flatpak run io.github.streetpea.Chiaki4deck", 0) == 0) {
+        filePath.append(getenv("HOME"));
```
nowrep, line 160: This should always save it in `$HOME/.config/Chiaki` and then you don't need special case for flatpak.

jbaiter:
Tend to agree on the whole artwork thing, we should provide a sensible default for the artwork and leave the customizations to the users, there are Decky plugins etc for that. It's a lot of code for a tiny feature.

The scripts themselves precede the gamepad/touch-friendly QML UI and are acceptable from that perspective, but maybe we should reconsider the idea once it's merged. Maybe a CLI subcommand to trigger the wakeup along with the connection? Then we could skip the script and simply launch the flatpak with custom arguments?

nowrep:
I'd say we should provide easy install method, something like what EmuDeck does - you download `.desktop` file, run it and it downloads install script and does the entire install. And if we have this, we should add the Steam entry as last install step automatically.

That would mean we only need to add new cli flag to (un)install Steam entry.

Nikorag:
Thanks for the feedback. As a first pass until we get direct connection remote/local from the CLI I'll clean the code up and move the artwork section to a CMAKE flag. Will update the PR when I'm done

gui/src/shortcutdialog.cpp
```diff
@@ -0,0 +1,311 @@
...
+    paramMap["wait_timeout"] = "35";
+    paramMap["home_addr"] = displayServer->GetHostAddr().toStdString();
+    paramMap["away_addr"] = external_dns_edit->text().toStdString();
+    paramMap["regist_key"] = QString::fromUtf8(displayServer->registered_host.GetRPRegistKey()).toStdString();
+    paramMap["login_passcode"] = passcode_edit->text().toStdString();
+    paramMap["mode"] = mode_combo_box->currentText().toStdString();
+    paramMap["server_nickname"] = displayServer->discovered ? displayServer->discovery_host.host_name.toStdString() : displayServer->registered_host.GetServerNickname().toStdString();
+    paramMap["executable"] = getExecutable();
+    std::stringstream escapeCommandStream;
+    escapeCommandStream << "printf %q " << "\"" << paramMap["server_nickname"] << "\"";
```
nowrep, line 125: Why would you ever need shell for this?

gui/src/shortcutdialog.cpp
```diff
@@ -0,0 +1,311 @@
...
+        QTextStream out(&outputFile);
+        out << QString::fromStdString(fileText);
+
+        // Close the file
+        outputFile.close();
+    }
+
+    // Construct the shell command to make the file executable
+    std::stringstream chmodCommandStream;
+    chmodCommandStream << "chmod +x " << "\"" << filePath << "\"";
```
nowrep, line 193: Same here

gui/src/mainwindow.cpp
```diff
@@ -354,6 +363,16 @@ void MainWindow::UpdateDisplayServers()
 	UpdateServerWidgets();
 }
 
+void MainWindow::OpenAddToSteamDialog(std::string nickname) {
+	QList<DisplayServer>::iterator it;
+	for (it = display_servers.begin(); it != display_servers.end(); ++it) {
+		if (it->registered_host.GetServerNickname() == QString::fromStdString(nickname)) {
```
nowrep, line 369: Server nickname isn't unique.

gui/src/registdialog.cpp
```diff
@@ -244,7 +244,15 @@ void RegistExecuteDialog::Success(RegisteredHost host)
 
 	settings->AddRegisteredHost(host);
 
+#if defined(__linux) || (defined(__APPLE__) && defined(CHIAKI_ENABLE_CLI))
```
nowrep, line 247: There is `Q_OS_LINUX` and `Q_OS_MACOS`.

gui/src/registdialog.cpp
```diff
@@ -244,7 +244,15 @@ void RegistExecuteDialog::Success(RegisteredHost host)
...
+	int addToSteam = QMessageBox::question(this, tr("Console registered"), tr("The Console %1 with ID %2 has been successfully registered! Would you like to add it to steam?").arg(host.GetServerNickname(), host.GetServerMAC().ToString()), QMessageBox::Yes, QMessageBox::No);
```
nowrep, line 248: `auto addToSteam = ...`

gui/src/serveritemwidget.cpp
```diff
@@ -40,6 +40,12 @@ ServerItemWidget::ServerItemWidget(QWidget *parent) : QFrame(parent)
 	addAction(wake_action);
 	connect(wake_action, &QAction::triggered, this, [this]{ emit WakeTriggered(); });
 
+	#if defined(__linux) || (defined(__APPLE__) && defined(CHIAKI_ENABLE_CLI))
+		create_shortcut_action = new QAction(tr("Add to Steam"), this);
+		addAction(create_shortcut_action);
+		connect(create_shortcut_action, &QAction::triggered, this, [this]{ emit CreateShortcutTriggered(); });
```
nowrep, line 46: `connect(create_shortcut_action, &QAction::triggered, this, &ServerItemWidget::CreateShortcutTriggered);`

gui/src/shortcutdialog.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+    //Screen Mode combo
+    static const QList<QPair<ChiakiScreenModePreset, const char *>> mode_strings = {
+        { CHIAKI_MODE_STRETCH, "stretch" },
+        { CHIAKI_MODE_ZOOM, "zoom" },
+        { CHIAKI_MODE_FULLSCREEN, "fullscreen" }
+    };
+    for(const auto &p : mode_strings)
+    {
+        Ui::ShortcutDialog::mode_combo_box->addItem(tr(p.second), p.first);
```
nowrep, line 39: This won't pick up the translations.

gui/src/shortcutdialog.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+    chiakiFlatpakDir.append(getenv("HOME"));
+    chiakiFlatpakDir.append("/.var/app/io.github.streetpea.Chiaki4deck");
+    //Installed via Flatpak
+    if (access(chiakiFlatpakDir.c_str(), 0) == 0) {
+        return "flatpak run io.github.streetpea.Chiaki4deck";
+    }
+    else {
+        char buffer[PATH_MAX];
+        ssize_t len = readlink("/proc/self/exe", buffer, sizeof(buffer) - 1);
```
nowrep, line 99: `QFileInfo::readSymLink()`

gui/src/steamgriddbapi.cpp
```diff
@@ -0,0 +1,88 @@
...
+    QJsonArray resultData = resultObject.value("data").toArray();
+
+
+    for (QJsonArray::iterator it = resultData.begin(); it != resultData.end(); ++it) {
+        QJsonObject gameObject = it->toObject();
+        QString imageUrl = gameObject.value("url").toString();
+        // Find the first occurrence of "\\/"
+        size_t found = imageUrl.toStdString().find("\\/");
+
+        // Iterate and replace all occurrences of "\\/"
```
nowrep, line 79: This shouldn't be needed to do manually, JSON parser should handle it (unless the contents of `url` really are escaped for some weird reason).

gui/src/steamshortcutparser.cpp
```diff
@@ -0,0 +1,660 @@
...
+
+    std::vector<char> fileHeader = { 0x00, 0x73, 0x68, 0x6F, 0x72, 0x74, 0x63, 0x75, 0x74, 0x73, 0x00 };
+    std::string controller_layout_workshop_id = "3049833406";
+
+    // Callback function to write data to a file
+    size_t WriteCallback(void* contents, size_t size, size_t nmemb, FILE* stream) {
+        return fwrite(contents, size, nmemb, stream);
+    }
+
+    void createDirectories(ChiakiLog* log, const std::string& path) {
```
nowrep, line 36: `QDir().mkpath()`

gui/src/steamshortcutparser.cpp
```diff
@@ -0,0 +1,660 @@
...
+                if (mkdir(subPath.c_str(), 0777) != 0) {
+                    CHIAKI_LOGE(log, "Error creating directory: %s", subPath.c_str());
+                    return;
+                }
+                CHIAKI_LOGI(log, "Created directory: %s", subPath.c_str());
+            }
+        }
+    }
+
+    void copyFile(ChiakiLog* log, const std::string& sourcePath, const std::string& destinationPath) {
```
nowrep, line 54: `QFile::copy()`

gui/src/steamshortcutparser.cpp
```diff
@@ -0,0 +1,660 @@
...
+
+        // Close both files
+        sourceFile.close();
+        destinationFile.close();
+
+        CHIAKI_LOGI(log, "File copied successfully from %s to %s", sourcePath.c_str(), destinationPath.c_str());
+    }
+
+    // Function to download a file from a URL to a destination directory
+    bool downloadFile(ChiakiLog* log, const char* url, const char* destPath) {
```
nowrep, line 83: The url should be `QUrl` and then you have `QUrl::isLocalFile`. But if you make real local urls `QUrl::fromLocalFile` then QNetworkAccessManager will handle them just fine same as http urls so you don't need to special-case it.

gui/src/steamshortcutparser.cpp
```diff
@@ -0,0 +1,660 @@
...
+        // Generate preliminary ID
+        uint64_t preliminaryId = generatePreliminaryId(exe, appname);
+
+        // Shift right by 32 bits and subtract 0x100000000
+        uint32_t shortcutId = static_cast<uint32_t>((preliminaryId >> 32) - 0x100000000);
+
+        return shortcutId;
+    }
+
+    std::string getValueFromMap(const std::map<std::string, std::string>& myMap, const std::string& key) {
```
nowrep, line 213: Or just use QHash/QMap and then you just do `map.value(key)`.

gui/src/steamshortcutparser.cpp
```diff
@@ -0,0 +1,660 @@
...
+        auto it = myMap.find(key);
+        return (it != myMap.end()) ? it->second : ""; // Return the value or an empty string if the key is not found
+    }
+
+    bool containsString(const std::vector<std::string>& stringArray, const std::string& targetString) {
```
nowrep, line 218: Don't really need helper function for this. Also it's shorter if you use QString instead.

gui/src/steamshortcutparser.cpp
```diff
@@ -0,0 +1,660 @@
...
+        auto it = std::find(stringArray.begin(), stringArray.end(), targetString);
+        return it != stringArray.end();
+    }
+
+    bool compareFirstElements(int size, const std::vector<char>& vector1, const std::vector<char>& vector2) {
+        // Check if both vectors have at least 11 elements
```
nowrep, line 224: This function isn't doing what this comment says.

gui/src/steamshortcutparser.cpp
```diff
@@ -0,0 +1,660 @@
...
+    }
+
+    void updateControllerConfig(ChiakiLog* log, std::string titleName) {
+        std::string controllerFile = getSteamBaseDir();
+        std::string userId = getMostRecentUser(log);
+        controllerFile.append("/steamapps/common/Steam Controller Configs/");
+        controllerFile.append(userId);
+        controllerFile.append("/config/configset_controller_neptune.vdf");
+
+        bool fileExists;
```
nowrep, line 615: `QFileInfo::exists`

gui/src/shortcutdialog.cpp
Nikorag: @streetpea this was in your PR, was there a reason for doing this in shell rather than native C?

Nikorag:
I've tidied up the code, and in doing so I believe I've addressed all the comments except for:

- Finding the executable. I don't have a linux system to test this on and in Apple's infiinite wisdom they've made it nigh-on impossible to virtualise one on an M1 mac. I require assistance with this. However, given the original script the template is based on had the flatpak name hardcoded in, we're not any worse off
- Translations. I wasn't aware Chiaki4Deck had translations?
- Pointless lambda in serveritem. This is in-keeping with the existing implementation of the other context menu items
- `Q_OS_LINUX` and `Q_OD_MAC`. The way it's being done is in-keeping with how it's achieved elsewhere in the codebase
- Shell used to remove special characters from the server nickname. I'm unsure why it was done this way. @streetpea raised it in a PR and I trust them!
- Initiating the dialog after registration uses non-unique nickname. Looking for tips here as I'm unsure how else to achieve this

streetpea:
I just used bash because that's how it was already done in the script and the script already relies on bash. Having said that, it would be preferable to write a function do this in c++ instead.

PR #114, opened 2023-12-12
Retrieve PSN Account ID from WebView

Nikorag:
Me Again!

I was reading all the praise for your new HDR feature on reddit and noticed that lots of people are put off by the steps needed to retrieve the PSN Account ID. So I built in a WebEngineView to automate the process and populate the dialog with the result.

~~Only issue is that I can't seem to work out how to return focus to the dialog once the WebView closes.~~

Looking for any input


jbaiter:
Great work! Do you think that instead of pulling in an additional dependency, we could use the Qt `QNetworkManager` to do the HTTP requests?

Nikorag:
I can take a look tomorrow at using a native QT library rather than curl. I'd have to apply the same change to the "Add to Steam" PR too.

Thanks for the feedback

Nikorag:
Hi @jbaiter 

I've removed Curl and used the `QNetworkAccessManager` instead

Thanks!

streetpea:
Hi @Nikorag I tried this out and it works nicely :+1:  I noticed 2 things that I would like your thoughts on:

1. I get a bunch of error messages despite it working (Is this avoidable?)
  ```
  [4409:7:1216/081009.433119:ERROR:command_buffer_proxy_impl.cc(141)] ContextResult::kTransientFailure: Failed to send GpuChannelMsg_CreateCommandBuffer.
  js: Refused to get unsafe header "X-Psn-Request-Id"
  js: Refused to get unsafe header "X-Psn-Correlation-Id"
  [4368:4402:1216/081011.387700:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081012.001799:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081012.013175:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081013.360315:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081013.367250:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081014.042786:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081014.051375:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081033.004201:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081033.041534:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081115.019249:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  [4368:4402:1216/081115.096823:ERROR:context_group.cc(146)] ContextResult::kFatalFailure: WebGL1 blocklisted
  js: Refused to get unsafe header "X-Psn-Request-Id"
  js: Refused to get unsafe header "X-Psn-Correlation-Id"
  js: Refused to get unsafe header "X-Psn-Request-Id"
  js: Refused to get unsafe header "X-Psn-Correlation-Id"
  ```
2. It looks like you force login to get psn account id. I think there should still be an option to directly enter an account id retrieved from somewhere else (such as https://psn.flipscreen.games/) for people that don't want to login to their psn account on Chiaki4deck with logging in being the default option and directly entering account id being a fallback option (maybe small line below button saying do you want to directly enter id instead of logging in to psn account or something like that)

Nikorag:
Hi!

Thanks for looking at this one too! I actually had the button just get the ID from the webview rather than replacing the text box originally but thought it was cleaner to replace it. Happy to change it back for people who'd prefer to enter it themselves. As for the warnings, which OS are you testing on? I'm using MacOS Ventura and I see the JS console errors for the headers but the GL errors are new to me. I'll do some reading to see if we should be concerned or if we can just silence them and move on.

Thanks again

Nikorag:
I think we can disable WebGL for the web engine view as it's not necessary: [look at WebGLEnabled here](https://doc.qt.io/qt-6/qwebenginesettings.html)

streetpea:
I tested on Manjaro Linux (that's what Valve recommended for Steam Deck dev). I can also try on steam deck via the flatpak and see if I also get this errors. As for the text box I agree that it looks better the way you have it but since I know some people will want to retrieve it manually (via making their id public and retrieving it vs. entering psn login credentials) and sometimes the page (if you're using the same one as the python script) gives a timeout error, I think it's important to give some sort of option for manual entry.

Nikorag:
Absolutely fair point. I'll take a look at it tomorrow. Sadly I only have an M1 Mac and virtualisation on x64 operating systems just doesn't seem to work so I really appreciate your testing on Linux for me. Out of interest does the project have a discord at all? Would be good to chat while working through these things to save me spamming you with PRs and comments.

streetpea:
We don't have a discord channel though I could create one if you think it would be helpful. You could also dm me on Reddit if you want. Regarding building on Linux you can use the dev instructions here to build it on your deck https://streetpea.github.io/chiaki4deck/diy/steamdeckdev/

That will work for your other pr already. for your current pr the flatpak is missing the qwebengine dep but I'll add it and post a new version you will be able to use and update the instructions shortly. (I'll let you know when it's ready for you to use)

Nikorag:
I've implemented your suggestion of allowing manual entry of the PSN account id. I also changed the window title to be the current URL in case anyone is worried about phishing. I've disabled webgl which should remove some of the errors but the unsafe header ones are legit. These are debug headers which Sony are probably using for analytics.

I haven't looked too far into it but if phishing is a real concern we could look to check the SSL certificate of each page before adding the lock to the window title bar. Let me know your thoughts.

streetpea:
You can now use this: https://streetpea.github.io/chiaki4deck/diy/steamdeckdev/ on your Steam Deck to build and test on there (just make sure to turn off vaapi) for the builds in there.

streetpea:
Also I created a discord here: https://discord.gg/YP8jZMxn if you prefer to chat there vs. GitHub

streetpea:
Based on your update and testing this out I think this is good to merge now. Are there any other updates you want to make to it @Nikorag?

Nikorag:
Nope. If you're happy, I'm happy!

streetpea:
btw if you also think it's easier to chat via discord vs. GitHub you can join the discord here: https://discord.gg/YP8jZMxn @jbaiter @nowrep

gui/include/jsonutils.h
```diff
@@ -0,0 +1,110 @@
...
+    private:
+        static std::string generateBasicAuthHeader(std::string username, std::string password) {
+            QString q_username = QString::fromStdString(username);
+            QString q_password = QString::fromStdString(password);
+            QString combined = QString("%1:%2").arg(QString(q_username)).arg(QString(q_password));
+            QString authHeader = "Basic " + combined.toUtf8().toBase64();
+            return authHeader.toStdString();
+        }
+
+        static std::string getJsonValueforAttribute(std::string jsonResponse, std::string jsonAttribute) {
```
nowrep, line 37: All of this should use `QJsonDocument`

gui/include/jsonutils.h
```diff
@@ -0,0 +1,110 @@
...
+            QNetworkReply *reply;
+            if (body.empty()) {
+                reply = manager.get(request);
+            } else {
+                QByteArray postData = QString::fromStdString(body).toUtf8();
+                reply = manager.post(request, postData);
+            }
+
+            // Setup a QEventLoop to wait for the request to finish
+            QEventLoop loop;
```
nowrep, line 90: Please don't spin another event loop unless necessary.

gui/include/psnaccountid.h
```diff
@@ -0,0 +1,89 @@
...
+            for (int i = 0; i < num_bytes; ++i) {
+                result[i] = static_cast<unsigned char>(number & 0xFF);
+                number >>= 8;
+            }
+        }
+
+        return result;
+    }
+
+    static std::string base64_encode(const std::vector<unsigned char>& input) {
```
nowrep, line 63: Could it use `QByteArray::toBase64` instead?

gui/src/mainwindow.cpp
```diff
@@ -274,7 +274,13 @@ void MainWindow::ServerItemWidgetTriggered()
 	else
 	{
 		RegistDialog regist_dialog(settings, server.GetHostAddr(), this);
-		int r = regist_dialog.exec();
+		regist_dialog.show();
+
+		QEventLoop loop;
```
nowrep, line 279: Why?

Nikorag: Without this loop the dialog closes immediately. It used to use dialog.exec() but this would lock focus and it would be impossible to view the login window as it was always behind the registration dialog. Pointing this out has actually  helped me to discover an issue with another change in this PR: `SettingsDialog::RegisterNewHost` I altered to use the show() logic but without the loop in oreder to allow switching focus to the WebView and the dialog never shows.

Do you have any suggestions for getting around this without loops? I tried using a lambda to remove the loop like below:

```		RegistDialog regist_dialog(settings, server.GetHostAddr(), this);
		connect(&regist_dialog, &QDialog::finished, this, [&regist_dialog, &server, this](int result) {
			if(result == QDialog::Accepted && !server.discovered) // success
			{
				ManualHost manual_host = server.manual_host;
				manual_host.Register(regist_dialog.GetRegisteredHost());
				settings->SetManualHost(manual_host);
			}
		});
		regist_dialog.show();
```

but without the loop itself the dialog just closed immediately.

Nikorag: Ignore this. I worked out if i mage the web engine inside a Dialog I can take focus from an the exec dialog beneath

gui/include/jsonutils.h
Nikorag: Are you able to assist in removing this loop? JsonUtils is a static class so without instantiating it I'm unsure how we could build this in an async way using slots and signals.

Nikorag:
I've addressed each of @nowrep 's comments except the QEventLoop within the static JsonUtils. Not sure where to go with that one.

gui/include/jsonutils.h
nowrep: This still should be changed to qjson.

hvmzx:
> you can join the discord here: https://discord.gg/YP8jZMxn

It would be cool to have a discord for the whole community, it's easier for people to talk and contribute to the app while also having more "support" in case there is a bug.

PR #121, opened 2023-12-16
Error message + disconnect dialog as overlay

nowrep:


streetpea:
I like this, it's much better UX then the current dialog message.

PR #122, opened 2023-12-16
libplacebo swapchain reuse

nowrep:
This seems to be a better solution.


streetpea:
Am I right in thinking that releasing this can wait until the next regularly scheduled release since the hot fix you made before has already been applied?

nowrep:
I reproduced similar issue also on stable. But I think it'll be better to give this time to properly test it before making a new release.

streetpea:
I'll merge this but wait until the next release to release it.

PR #127, opened 2023-12-18
StreamConnection: Fix crash when closing stream connection before it …

nowrep:
…starts


streetpea:
Is this ready to be merged or just the start of a bigger fix?

nowrep:
Ready. It shouldn't crash now when pressing Ctrl+Q while starting session.

PR #145, opened 2024-01-03
Reconnect session on resume from sleep

nowrep:
Stop session before going to sleep and try to reconnect on resume from sleep.

### DIFF
--- a/gui/CMakeLists.txt
+++ b/gui/CMakeLists.txt
@@ -3,6 +3,9 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_AUTORCC ON)
 find_package(Qt6 REQUIRED COMPONENTS Core Gui Concurrent Svg)
+if(UNIX AND NOT APPLE)
+    find_package(Qt6 REQUIRED COMPONENTS DBus)
+endif()
 if(APPLE)
 	find_package(Qt6 REQUIRED COMPONENTS MacExtras)
 endif()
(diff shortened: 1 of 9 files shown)


streetpea:
another thing to note is that you can get a remote play is in use error if you reconnect too quickly (before the remote play server has a chance to shut down ... when I tested this out I noticed that and adding `--system-talk-name=org.freedesktop.login1` seems sufficient instead of the whole system bus (at least turning my console to sleep and turning back on resulted in the console reconnecting).

nowrep:
> another thing to note is that you can get a remote play is in use error if you reconnect too quickly

Yeah that already happens when just trying to connect right after disconnecting. But for me that 10s is enough for it to succeed eventually (maybe if needed it can be increased to 15s?). Also it should definitely not be an issue when putting deck to sleep as the session is stopped before going to sleep and then it takes additional time to finish sleep -> resume before it stars to connect to console.

streetpea:
If the console goes to sleep after the Steam Deck sleeps (for example power saving settings on the PS5 are on and a certain amount of time has passed causing the PS5 to enter sleep mode [can manually put the ps5 in sleep mode for testing purposes]) this will fail. May want to check if the PS5 is ready and if not wakeup first.

nowrep:
> ... this will fail

And that's fine, there is multiple other reasons why it can fail. 

> May want to check if the PS5 is ready and if not wakeup first.

Can't really do that because it requires discovery response which may not be ready at the time for various reasons. The wakeup packet could be unconditionally sent in `StreamSession::Start()` but not sure if good idea.

nowrep:
Wakeup works now.

streetpea:
The discovery manager uses broadcast and so it only seems to work for local connections so this version of wakeup will likely only work for local connections not remote. That's fine as a limitation though.

PR #152, opened 2024-01-06
VideoReceiver: Parse H265 bitstream and skip decoding corrupted frames

nowrep:
Keep track of reference frames and skip decoding frames with missing reference frames.

This gives much better results as now we only decode and render frames that *should* be decodable without any errors.

I've pulled in the bitstream reader code from Mesa, because it's proven to be working well and as far as I understand there should be no issues with licence.


streetpea:
this doesn't compile with libplacebo off like `cmake -DCHIAKI_ENABLE_PLACEBO=false ..` it looks like the opengl stuff needs to be updated.

nowrep:
Fixed

PR #154, opened 2024-01-07
AppImage update for Qt 6

nowrep:
* fixed VAAPI
* also fixed Vulkan, but SteamOS has old drivers
* fixed haptics


streetpea:
running results in cmake not found error

nowrep:
Should work now

PR #157, opened 2024-01-08
 CongestionControl: Disable reporting packet loss

nowrep:


streetpea:
In a situation where the bitrate is too high for the connection wouldn't tanking the quality be preferable to keeping an unsustainable bitrate ... also are you seeing the target bitrate decreasing (you should see these reported in verbose log messages) maybe just resending connection quality protobufs with the higher target bitrate would fix this?...

nowrep:
> In a situation where the bitrate is too high for the connection wouldn't tanking the quality be preferable to keeping an unsustainable bitrate

It wouldn't, that's why there is bitrate settings and in this case user should manually lower it. The default values are low enough that it should be always good, and in case it isn't then the streaming experience won't be good anyway.

> also are you seeing the target bitrate decreasing

Yes, I am. Both the target bitrate sent from console and the observed calculated bitstream (displayed in stream menu) decreases permanently.

You can test it easily by yourself, simply go to Settings -> Internet maybe couple times and you'll be stuck with extremely low bitrate until restarting the session. So until we can fix it somehow, it's better to just disable this feature completely imo.

> maybe just resending connection quality protobufs with the higher target bitrate would fix this

I haven't found any message there that would affect bitrate in any way. The only time the client sends bitrate to console is in the launchspec and it's just a bitrate "hint".

PR #159, opened 2024-01-08
Autoconnect to a console by nickname from the CLI

Nikorag:
This should replace the need to generate launcher scripts at all.

The implementation of remote access isn't fantastic, it uses a second process to start the chiaki CLI and run discover and wakeup as the DiscoveryManager class in the gui doesn't have the option of querying a specific host like the CLI does.

Off the back of this I'll make changes to the Add to Steam PR to simplify the new shortcut so it doesn't rely on creating a shell script and just adds Chiaki straight in.


gui/include/autoconnecthelper.h
```diff
@@ -0,0 +1,155 @@
+#ifndef SSIDHELPER_H
```
jbaiter, line 1: Why do this by shelling out? The Chiaki subcommands themselves (or rather the code that implements them) can just be called via C++.

I'm sure there are OS-specific APIs that can be used to get the SSID of the currently connected WiFi AP as well (NetworkManager's dbus API has the info e.g.), but it's probably a lot more code to make it truly cross-platform, so it's probably ok to shell out for that.

jbaiter (changes requested):

gui/include/autoconnecthelper.h
Nikorag: I looked into calling the shell commands but they just return an int status code of the command result rather than any objects I can infer the console's from. I think, from my rudimentary understanding, they all handle the discovery asynchronously via callbacks. Though now I've written that down I'm not sure it's right. I think it uses callbacks but it's all synchronous as the terminal waits for a response. Let me take a look when I get a minute and see if I can sort it.

Nikorag:
As discussed on Discord, I checked for whether we are able to initiate the stream directly without calling the discover and wake commands and this wasn't successful so I've left them in place. The gen-launcher script doesn't do this either so I was skeptical.

The talk of removing the discover step entirely from the local pipeline isn't possible as Registered Hosts do not save a static IP against them. I assume this is so if their DHCP expires and they're re-assigned, you wouldn't need to re-register.

I have however replaced the shell calls for discover with a new method in the DiscoveryManager which can query a specific host rather than via broadcast. As I mentioned in my response to @jbaiter, the way the discovery service works is sort-of asynchronous in that I needed to provide a static callback. I pass a pointer through to the callback of the instance of DiscoveryManage and update a `state` variable when I get the response. It's not thread-safe at all so could do with re-work however It looked like some fundamentals would need to be altered for what's a tiny QOL improvement.

Nikorag:
I looked at resolving the conflicts and I can see an alternative method of autoconnect has been merged. If this is how chiaki4deck handles autoconnect neither of my PRs is necessary anymore as you wouldn't have separate launchers. Would you like me to close them?

streetpea:
I think adding to steam is still useful because you can add controller layout, etc. in a simpler way ... the current auto connect still seems to be a work in progress but this form of auto-connect is probably unnecessary

Issue #196, opened 2024-01-17
Issue with macOS arm64 build


outcoldman:
It is possible that arm64 builds [must be signed](https://eclecticlight.co/2020/08/22/apple-silicon-macs-will-require-signed-code/). Seems like Intel build is not signed, and macOS allows to execute it anyway. In case of arm64 the user will receive that app is damaged and suggest moving it to Trash. 

Workaround:

1. Remove quarantine flag (that it was downloaded from the internet). Run in terminal `xattr -dr com.apple.quarantine Chiaki4deck.app` in the folder where the `Chiaki4deck.app` is.
2. Sign it locally. To do that, you need a `codesign` tool, which might be available after you install Command Line Tools or Xcode. Maybe it is actually available without. Run in terminal `codesign --force --deep --sign - ./Chiaki4deck.app` in the folder where `Chiaki4deck.app` is. 

From that article above:

> Developers without an Apple certificate can just sign their apps with a 3rd-party certificate (non-Apple) or ad-hoc, and offer that version as a download. You’ll still need to get around Gatekeeper (right-click-Open, remove quarantine XA etc.), but that’s no different than it is today. As for the rest: as long as the bundle’s signature is and remains intact, the app should work & continue to work on ARM-BS+

PR #197, opened 2024-01-17
Add ad-hoc signing for the arm64 build

outcoldman:
Trying to add ad-hoc signing to fix issue #196 

cc @nowrep


nowrep:
Does this actually helps? My understanding is that you need to pay for developer license, and using self-signed cert will just show the same untrusted warning.

outcoldman:
@nowrep yeah, it still will show untrusted version, but right now for the arm64 apps, it would not even launch it. It will say that it is damaged and only option will give you to move it to trash. See #196

So this "fix" is only making arm64 build to make it the same as Intel experience. Not better. Just the same. As currently you cannot even launch this app on Apple silicon.

PR #233, opened 2024-01-27
Add UDP as needed port-forward

Stals:
When connecting remotely, based on Wireshark it sends both TCP & UDP to 9295. Only after updating port forwarding to forward both TCP and UDP it started to connect
It seems the 9295 UDP is only used during registration, but anyway, if it is not port-forwarded, then it isn't possible to add the PS5

image

Although I'm not a 100% sure if all 9295-9297 should be TCP & UDP, but 9295 definitely seems that way.

### DIFF
--- a/docs/setup/remoteconnection.md
+++ b/docs/setup/remoteconnection.md
@@ -16,7 +16,7 @@ Forward the ports for your console on your router following [this port forwardin
 
     | Port | Connection Type |
     | ---- | --------------- |
-    | 9295 | TCP             |
+    | 9295 | TCP & UDP       |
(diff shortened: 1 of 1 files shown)


streetpea:
You should be able to add the PS5 after registering locally following the instructions in the guide using the + button. Only local registration is supported not remote registration.

Stals:
But the change to enable both TCP and UDP for these ports actually allowed me to do a remote registration; due to the specific router model and also Steam Deck and PS5 being in different countries, it wasn't really possible to do a local registration, but this worked.

PR #238, opened 2024-01-31
Add CUDA and QSV to allowed decoder list

Sn0wCrack:
In the original Chiaki (and most of the Qt6 ports I've used), the entirety of the decoder list was added to this dropdown box. This allowed me to use CUDA and QSV on systems where I had access to these.

I've added CUDA and QSV to the allowed decoders list for all Operating Systems, as far as I'm aware these should both be available on Linux and Windows at the very least.

I've tested both CUDA and QSV by manually editing my saved registry values on Windows and both seem to work fine, I don't have easy access to a Linux device to test on there.

I'm happy to move these into the Windows and Linux blocks specifically to prevent macOS users from selecting it.

There are other decoders that could be enabled, such as DVXA2 (Windows-only) and OpenCL, however I'm unaware of either of these have any practical purpose now that Vulkan, VA-API and D3D11VA exist and are decent on most systems.

### DIFF
--- a/gui/src/qmlsettings.cpp
+++ b/gui/src/qmlsettings.cpp
@@ -241,6 +241,8 @@ QStringList QmlSettings::availableDecoders() const
 {
     static QSet<QString> allowed = {
         "vulkan",
+        "cuda",
+        "qsv",
 #if defined(Q_OS_LINUX)
         "vaapi",
 #elif defined(Q_OS_MACOS)


nowrep:
Is there a reason why you can't use Vulkan? Only Vulkan and VAAPI have zero-copy, so ideally we should only use those.

In any case, it would be better to check the GPU vendor and filter based on it (also check Vulkan video support).

Sn0wCrack:
> Is there a reason why you can't use Vulkan? Only Vulkan and VAAPI have zero-copy, so ideally we should only use those.

Mainly related to my bug report about Vulkan crashing, setting the value to cuda in the registry settings seems to run well and uses my GPU for video decode without issue.

> In any case, it would be better to check the GPU vendor and filter based on it (also check Vulkan video support).

Fair point, I'll look into trying to do this.

nowrep:
I'd prefer fixing Vulkan instead.

Also what's wrong with using d3d11va?

Sn0wCrack:
d3d11va does seem to work fine for my use cases so far but will have to get a full session out of it to see how it goes.
From my brief testing, d3d11va does use more GPU utilization than when using CUDA at idle.

policeman0077:
> > Is there a reason why you can't use Vulkan? Only Vulkan and VAAPI have zero-copy, so ideally we should only use those.
> 
> Mainly related to my bug report about Vulkan crashing, setting the value to cuda in the registry settings seems to run well and uses my GPU for video decode without issue.
> 
> > In any case, it would be better to check the GPU vendor and filter based on it (also check Vulkan video support).
> 
> Fair point, I'll look into trying to do this.

which registry I should use to set it to cuda

PR #247, opened 2024-02-04
Refresh Windows MSVC build

r52:
This PR brings the native Windows MSVC build up to par with the MSYS2 builds. Totally unnecessary, but nice to have for educational purposes.

Everything works as far as I can tell, including both hardware decoders.

Some caveats/notes for future reference:
- The build script now uses vcpkg to automatically bring in most dependencies with a few exceptions: ffmpeg, shaderc, spirv-cross, and libplacebo
- ffmpeg and shaderc are simply much faster and more convenient to just download the prebuilt binaries
- libplacebo must be built from scratch because it doesn't exist in the vcpkg repository and has no prebuilt binaries
- spirv-cross must be built from scratch because libplacebo requires spirv-cross-c-shared when building for d3d11 support and vcpkg only offers the static version
- The meson config file (meson.ini) used by the libplacebo build needs to be in the project root because it only accepts absolute paths, so it must be in the project root in order to see vcpkg installed dependencies
- libplacebo produces an invalid pkgconfig file for MSVC due to binary naming discrepancies, so it needs to be patched
- Fixed ffmpeg linking on machines without system-level ffmpeg installs

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -1,4 +1,3 @@
-# Adpated from nlz242/chiaki-qt6
 name: Build Chiaki4deck Windows (VC)
 
 on:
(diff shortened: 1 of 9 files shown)


r52:
Changed to acquire vcpkg on the fly

Issue #263, opened 2024-03-03
[BUG] Application crashes on startup (SD OLED)


andreylugovtsov:
Hi!
I haver an oled model with the latest software updates available at the moment.
I've installed chiaki4deck via desktop's Store application.
It's crashing on launch. Icon would shortly show up in the taskbar and then disappear.
When `flatpak run --env=QT_DEBUG_PLUGINS=1 io.github.streetpea.Chiaki4deck` is executed, there is a normal prompt to determine PSN account id.
Meanwhile chiacki runs fine.
Tried a solution from https://github.com/streetpea/chiaki4deck/issues/112 `sudo flatpak override --nosocket=fallback-x11 --socket=x11 io.github.streetpea.Chiaki4deck`, but it didn't help.
Also tried reinstalling/rebooting - no luck.

brucetony:
Same error occurring for me, here is the output of running `flatpak run io.github.streetpea.Chiaki4deck`:
```bash
(deck@steamdeck ~)$ flatpak run io.github.streetpea.Chiaki4deck > ~/tmp/chiaki_error.txt
F: Not sharing "/esp" with sandbox: Ignoring blocking autofs path "/esp"
chiaki.gui: [libplacebo] Initialized libplacebo v7.342.0 (API v342)
chiaki.gui: [libplacebo] Spent 47.890 ms enumerating instance extensions
chiaki.gui: [libplacebo] Creating vulkan instance with extensions:
chiaki.gui: [libplacebo]     VK_KHR_get_physical_device_properties2
chiaki.gui: [libplacebo]     VK_KHR_surface
chiaki.gui: [libplacebo]     VK_EXT_swapchain_colorspace
chiaki.gui: [libplacebo]     VK_KHR_external_memory_capabilities
chiaki.gui: [libplacebo]     VK_KHR_external_semaphore_capabilities
chiaki.gui: [libplacebo]     VK_KHR_get_surface_capabilities2
chiaki.gui: [libplacebo]     VK_KHR_portability_enumeration
chiaki.gui: [libplacebo]     VK_KHR_xcb_surface
chiaki.gui: [libplacebo]     VK_KHR_surface
chiaki.gui: [libplacebo] Probing for vulkan devices:
chiaki.gui: [libplacebo]     GPU 0: AMD Radeon Graphics (RADV VANGOGH) v1.3.274 (integrated)
[...]

PR #261, opened 2024-03-02
Add missing QtQuick.Controls.Material imports

devusb:
In a Plasma 6 desktop (tested on NixOS), chiaki4deck fails to launch with errors like
```
chiaki.gui: Component not ready
 QList(qrc:/Main.qml:171:9: Type ManualHostDialog unavailable
            ManualHostDialog { }
            ^, qrc:/ManualHostDialog.qml:9:1: Type DialogView unavailable
    DialogView {
    ^, qrc:/DialogView.qml:72:17: Non-existent attached object
                    Material.roundedScale: Material.SmallScale
                    ^)
```
This seems to be due to missing imports of `QtQuick.Controls.Material` in a few spots -- `Material` is used but not explicitly imported, resulting in the errors. This PR adds the missing imports.

I am able to launch chiaki4deck successfully (and get into a session with a PS5) with these changes.

PR #264, opened 2024-03-03
Update DialogView.qml

Golden-z:
to fix a crash error reported in issue #263


brucetony:
#261 solves the issue this PR is trying to fix. I suggest going with #261 for LTS

Golden-z:
issue resolved by #261

PR #312, opened 2024-04-15
Enable don't fragment for macOS

shatyuka:
macOS has supported the "Don't Fragment" feature since Big Sur. Since we target at Qt 6.6, just simply add it into FreeBSD branch.

Previous behavior:
```
[I] Takion connecting (version 7)
[W] Don't fragment is not supported on this platform, MTU values may be incorrect.
[I] Takion sent init
```

Current behavior:
```
[I] Takion connecting (version 7)
[I] Takion enabled Don't Fragment Bit
[I] Takion sent init
```


streetpea:
thanks for letting me know, I'll look into this

PR #320, opened 2024-05-05
Remote Connection Using PSN for Setup

streetpea:
**Description**
This PR enables creating a remote connection using PSN as an initial go between to setup the connection via UDP holepunching to eschew the need to forward ports.

**Requirements**
- Console must be connected to PSN and updated to the latest firmware
- Console must be registered either locally or remotely via the traditional registration
- Home network must be compatible with UDP holepunching (not all networks are compatible)

**Notes**
- Console will show up as Remote Connection Via PSN with the console nickname
- Console will only show up for remote connection if console is not discovered locally (otherwise Chiaki4deck will use local connection)

**Acknowledgements**
Besides the code contributors included in the commits, this was also made possible by @grill2010 who contributed his knowledge and data from reverse engineering this feature for PSPlay.


streetpea:
Todo before release:
- [ ] Update Documentation for PSN Remote Connection
- [ ] Anything else?

PR #389, opened 2024-08-13
gui: Add new 'custom' quality preset for placebo

streetpea:
Will monitor the file `pl_render_params.conf` in the Chiaki config directory (~/Library/Preferences for MacOS see https://doc.qt.io/qt-6/qstandardpaths.html for other platforms) for custom rendering settings and apply them to the running stream. For available options see [1], the syntax is simply a set of `key=value` pairs, separated from each other with either whitespace, comma, semicolon, colon or newline characters.

[1] https://libplacebo.org/options/

### DIFF
--- a/gui/src/qmlmainwindow.cpp
+++ b/gui/src/qmlmainwindow.cpp
@@ -55,6 +55,13 @@ static QString shader_cache_path()
     return path;
 }
 
+
+static const char *render_params_path()
+{
+    static QString path = QStandardPaths::writableLocation(QStandardPaths::ConfigLocation) + "/Chiaki/pl_render_params.conf";
+    return qPrintable(path);
+}
+
 class RenderControl : public QQuickRenderControl
 {
 public:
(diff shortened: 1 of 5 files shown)


gui/src/qmlmainwindow.cpp
```diff
@@ -55,6 +55,13 @@ static QString shader_cache_path()
     return path;
 }
 
+
+static const char *render_params_path()
+{
+    static QString path = QStandardPaths::writableLocation(QStandardPaths::ConfigLocation) + "/Chiaki/pl_render_params.conf";
```
jbaiter, line 61: I'm not really sure about this one, it did correctly resolve to `~/.config/Chiaki` on Linux for me, but the Qt6 docs say that on Windows it resolves to `"C:/Users/<USER>/AppData/Local/<APPNAME>"`, i..e  the added `/Chiaki/` would not be neccessary in this case.
I don't remember why I didn't use `QStandardPaths::AppConfigLocation` 🤔

nowrep: Yeah this should be `QStandardPaths::AppConfigLocation`

streetpea:
@jbaiter I made a pull request to your branch adding the placebo options via the Settings (with a button to open the placebo settings when the custom renderer preset is selected)

PR #479, opened 2024-11-19
io: collate fork patches for rumble/hwdecode/resolution

xlanor:
This includes kkWong's patches for chiaki that I've applied against the main tree. commit author has been left as the original author.

I've rebuilt the docker environment with switch-dav1d bundled in, no need to re-dl on compile of chiaki. Please re-pull the docker image.

The changes are summarized:
- yellow color correction
- hardware acceleration
- resolution bump
- haptic rumble.

I reached out to kkWong over the associated email with the patches but received no reply. Given both forks are under AGPLv3, do we need to add attribution anywhere else?


xlanor:
Hold first, 1080p stream is bugging out after tests

xlanor:
Additional crash found on exit, probably not cleaning up properly.

PR #484, opened 2024-11-24
feat: add ability to lookup psn id via public profile

xlanor:
As a todo, I had to shut off SSL for this request to work. 

When entering an account name now, I am able to fetch the associated account ID and write to cfg file + update the other input list item.

gui file was also too long so I started putting it out into views/

Set +x bit on sh files too.

Note: I have to remove SSL for now. this is because on modified switches it is very common to run it in incognito mode, specifically  blank_prodinfo_sysmmc=1 in exosphere.ini to prevent it from writing logs to sysnand and getting banned.

How this will affect PS Remote Play capabilities, I have to investigate further.


streetpea:
As you alluded to I don't think remote connection via PSN will work without tls so will probably need a workaround to get that working.

xlanor:
> As you alluded to I don't think remote connection via PSN will work without tls so will probably need a workaround to get that working.

Yep. I looked over my switch setup. I suspect it may have been a legacy setting from when I first set it up.

I will reset it up again and see if tls works.

PR #522, opened 2025-01-07
Suppress Apple warning for unsigned app & update docs

kylerm42:
Recently saw this error when installing via Homebrew on macOS:
image

Stumbled across [another repo](https://github.com/nikitabobko/AeroSpace?tab=readme-ov-file#installation) that seems to have a way to suppress that warning. I've attempted to do the same here, but I'm not at all familiar with homebrew packages, so I can't guarantee I've done it right for this repo.

I've also attempted to update the documentation with a note about the warning message suppression. Maintainers will probably want to ensure it's using the language the want.

### DIFF
--- a/docs/setup/installation.md
+++ b/docs/setup/installation.md
@@ -39,6 +39,17 @@
         ```
         brew install --cask streetpea/streetpea/chiaki-ng
         ```
+    
+    > [!NOTE]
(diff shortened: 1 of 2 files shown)


streetpea:
You’ll want to make the pull request at https://github.com/streetpea/homebrew-streetpea

PR #541, opened 2025-02-01
Add additional categories

razzeee:
This is an attempt to help with grouping categories better. I've send MR like this to other game streaming projects.


streetpea (approved):
lgtm

PR #564, opened 2025-02-27
Display correct icon when running under Wayland on Linux

Sn0wCrack:
When running under Wayland, an Application ID needs to be provided for the program icon to appear.

In Flatpak distributions this is the ID on Flathub essentially.

Additionally this can be set in SDL in the new version 3 to have the icon display in certain Desktop environment's volume manager.

### DIFF
--- a/gui/src/main.cpp
+++ b/gui/src/main.cpp
@@ -96,7 +96,7 @@ int real_main(int argc, char *argv[])
 		return 1;
 	}
 
-    	SDL_SetHint(SDL_HINT_APP_NAME, "chiaki-ng");
+	SDL_SetHint(SDL_HINT_APP_NAME, "chiaki-ng");
 
 	if(SDL_Init(SDL_INIT_AUDIO) < 0)
 	{
--- a/gui/src/main.cpp
+++ b/gui/src/main.cpp
@@ -69,6 +69,9 @@ int real_main(int argc, char *argv[])
 	QGuiApplication::setApplicationDisplayName("chiaki-ng");
 	QGuiApplication::setDesktopFileName("chiaki-ng");
 
+	// TODO: When upgrading to SDL3, this also needs to be set on "SDL_HINT_APP_ID"
+	QApplication::setDesktopFileName("io.github.streetpea.Chiaki4deck");
+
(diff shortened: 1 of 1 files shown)


Sn0wCrack:
I'm happy to make that adjustment for this to only set when under a Flatpak environment, however this does go a bit beyond Flatpak.

While I did make this PR with the intention of fixing the icon while using the Flatpak under Wayland, the XDG Desktop File standards do state the name of the desktop file should essentially be your application ID: https://specifications.freedesktop.org/desktop-entry-spec/latest/file-naming.html

Currently QT under Wayland only supports setting the top level icon via the Desktop file as well as it does not support the top level icon Wayland protocol currently. Same goes for SDL 3, it only supports this via the Desktop file.

streetpea:
@Sn0wCrack from reading that it seems the desktopFileName is correct as-is as chiaki-ng as that is currently what the desktop file is (chiaki-ng.desktop) ... it only gets changed to io.github.streetpea.Chiaki4deck within the flatpak environment as everything is renamed according the appid in that environment.

streetpea:
so it seems it's already set correctly as chiaki-ng.

streetpea:
@Sn0wCrack it's chiaki-ng regularly but in the flatpak the desktop file gets renamed to the flatpak id (see https://docs.flatpak.org/en/latest/conventions.html#desktop-files) or in their words
```
Flatpak will export the following desktop filename patterns: FLATPAK_ID.desktop, $FLATPAK_ID.foo.desktop, $FLATPAK_ID-foo.desktop. Exported desktop files can be found in the applications subfolder of either $HOME/.local/share/flatpak/exports/share or /var/lib/flatpak/exports/share, depending on whether it’s a system or user install.
```
the desktop filename gets changed from chiaki-ng to io.github.streetpea.Chiaki4deck for the flatpak only. The name is already correctly set for non-flatpak, so you just need to add a case for the flatpak version.

streetpea:
I guess what I'm trying to say is I see what you're saying but if I change the desktop file name it will be annoying for users so I don't want to change it at this point. Since the flatpak creates its own desktop file after the name of the id I'll just use that one for the flatpak.

PR #565, opened 2025-03-01
Display correct icon when running under Flatpak on Wayland

Sn0wCrack:
There's no truly definitive way of determining you're in a Flatpak, but checking for the existence of `/.flatpak-info` seems to work well.

### DIFF
--- a/gui/src/main.cpp
+++ b/gui/src/main.cpp
@@ -96,7 +96,7 @@ int real_main(int argc, char *argv[])
 		return 1;
 	}
 
-    	SDL_SetHint(SDL_HINT_APP_NAME, "chiaki-ng");
+	SDL_SetHint(SDL_HINT_APP_NAME, "chiaki-ng");
 
 	if(SDL_Init(SDL_INIT_AUDIO) < 0)
 	{
--- a/gui/src/main.cpp
+++ b/gui/src/main.cpp
@@ -67,10 +67,13 @@ int real_main(int argc, char *argv[])
 	QGuiApplication::setApplicationName("Chiaki");
 	QGuiApplication::setApplicationVersion(CHIAKI_VERSION);
 	QGuiApplication::setApplicationDisplayName("chiaki-ng");
-	QGuiApplication::setDesktopFileName("chiaki-ng");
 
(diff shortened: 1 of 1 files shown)


streetpea:
a75d463 should fix this

streetpea:
I don't know if checking for FLATPAK_ID or flatpak-info is better I just already did it the FLATPAK_ID way before you made this pr.

streetpea:
also @Sn0wCrack do you know if it actually checks if it's a reverse dns or not? I'm guessing it just didn't work for the flatpak because it exports the desktop file to flatpak in the appid.desktop format so the chiaki-ng.desktop doesn't exist within the flatpak and thus it can't find the desktop file which then links to the icon but not sure if it actually needs to be a reverse dns or not.

Sn0wCrack:
> I don't know if checking for FLATPAK_ID or flatpak-info is better I just already did it the FLATPAK_ID way before you made this pr.

I believe both work just as well really, Flatpak doesn't offer a definitive way of checking your application running under a Flatpak annoyingly.

> also @Sn0wCrack do you know if it actually checks if it's a reverse dns or not? I'm guessing it just didn't work for the flatpak because it exports the desktop file to flatpak in the appid.desktop format so the chiaki-ng.desktop doesn't exist within the flatpak and thus it can't find the desktop file which then links to the icon but not sure if it actually needs to be a reverse dns or not. If it actually needs to be a reverse dns to show the icon for the appimage I guess I can change the desktop file name to be a reverse dns but it will just cause users to have double shortcuts for chiaki-ng

I do not believe it actually bothers validating this fact currently, this function simply instructs the Window Manager to search for a specific desktop file with the provided name.

The reverse DNS format is simply just the standard as defined under Free Desktop and is meant to align the Desktop File, Application ID and D-Bus bus name.
Some applications and environments may use this standard to attempt to infer things, such as the Application ID inferring the Desktop File name and D-Bus bus name, etc.

Once Qt and KDE support [xdg-toplevel-icon-v1](https://wayland.app/protocols/xdg-toplevel-icon-v1) needing to do this will be less of a necessity as well I hope.

PR #734, opened 2026-03-25
holepunch: expose configurable port guessing for NAT traversal

xlanor:
Expose configurable port guessing API for NAT traversal on NATs where the address is same but ports may get rewritten, if the flags are set to enabled via api, offer adjacent ports when local_port != stun_port and increment is 0.

Thanks to grill as well who I was discussing this with

The default for said flags should fall back to the default numbers (75/250) if this api is not set.

Also switched to `poll` on switch rather than `select` 

still going through final tests downstream, but this is what it looks like when implemented in client:

2026032523460300-7DCC42E2AF4C1BBE54BB71700F7161B6

### DIFF
--- a/lib/include/chiaki/remote/holepunch.h
+++ b/lib/include/chiaki/remote/holepunch.h
@@ -194,6 +194,34 @@ CHIAKI_EXPORT ChiakiHolepunchSession chiaki_holepunch_session_init(
     const char* psn_oauth2_token, ChiakiLog *log);
 
 /**
+ * Enable port guessing for NAT traversal.
+ *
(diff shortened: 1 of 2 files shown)


streetpea:
Note: I updated the apis a little bit decoupling the set ports and force port guessing and made it so the set ports also affects the port guessing that takes place in the old guessing path as well as the new force guessing path.

xlanor:
> Note: I updated the apis a little bit decoupling the set ports and force port guessing and made it so the set ports also affects the port guessing that takes place in the old guessing path as well as the new force guessing path.

Thanks! Will rebase downstream and adjust calls

PR #745, opened 2026-04-06
Properly check for broken versions of radv rather than disallowing Vulkan on SteamOS with AppImages

MapleHinata:
Added a check for known broken radv version if we're running on Van Gogh GPUs, rather than disallowing Vulkan HW Decoding on AppImages on SteamOS. 

This allows running Vulkan HW decoding with AppImages on SteamOS 3.7 or newer, and fixes falling back to vaapi on non-SteamOS distros (In case someone is running their Steam Deck/Xbox ROG Ally on Debian stable or something along those lines lol.) 

(And yes, i did test this, going back to an mesa version that old on SteamOS 3.8 was a bit of a pain but I *did* test this!)


MapleHinata:
I just noticed that previous fix for release versioning on Windows builds did not apply to PRs, so I took the liberty of fixing it in here too, I hope you don't mind.

streetpea:
@MapleHinata thanks for the pr and yeah of course that's fine

PR #793, opened 2026-08-04
Feature/dependabot

zetneteork:


streetpea:
I think this should be added to the weekly build instead of run seperately

PR #802, opened 2026-08-23
Basic AIA chain recovery for certificates

### LINKED ISSUE #796: what is going on? "something went wrong", im trying to re-sign back to psn to link chiaki for offline remote play. but i get this. everytime.
Image

### LINKED ISSUE #798: [BUG] Remote connection over PSN (outside network) doesn't work
**Describe the bug**
I tried connecting to my PS5 over PSN connection from another network (local network connection worked fine) but I kept getting this error in the logs, while the UI infinitely loads saying "Establishing connection with console over PSN". I tried it on different machines, with different Linux distributions, and different networks (even mobile hotspot), always the same result. I also tried it on the AppImage version, also the flatpak, same error.

**Debug Log**
```
[E] websocket_thread_func: Connecting to push notification WebSocket wss://44-234-162-2-pushcl.np.communication.playstation.net/np/pushNotification failed with CURL error SSL peer certificate or SSH remote key was not OK
```

**To Reproduce**
1. Open Chiaki while outside of the PS5 network
2. Click on Remote Connection via PSN
3. Infinite loading
4. Force kill the app and check logs to see the debug log attached.

**Expected behavior**
Remote PSN connection should work.

**Desktop (please complete the following information):**
 - OS: Tried on Fedora Linux 44 and Void Linux
 - Device: Laptop
 - chiaki-ng Version 1.10.0

### LINKED ISSUE #799: [BUG] "Establishing connection with console over PSN" hangs when trying to register local PS5
**Describe the bug**
chiaki-ng was working as normal (both local network connection and remote connection were successful) until, in an attempt to connect via remote connection, it suddenly hung at the "Establishing connection with console over PSN" screen. I have uninstalled and reinstalled chiaki-ng several times and logged out and back into PSN via token several times. Token login is successful from Desktop and Gaming Mode, but when trying to register the PS5 over local Wi-Fi, it hangs at the same "Establishing connection with console over PSN" screen. chiaki-ng also hangs when attempting to cancel establishing a connection over PSN. chiaki-ng is installed via the flatpak. 

chiaki-ng is also unable to connect to the PS5 via manually registering the console. That error reads:

> Register Console
> [I] Regist starting search
> [I] Regist sending search packet
> [E] Regist failed to send search: Invalid argument
> [E] Regist search failed

**Debug Log**
[chiaki_session_2026-08-21_21-57-14-542542.log](https://github.com/user-attachments/files/31326262/chiaki_session_2026-08-21_21-57-14-542542.log)

**To Reproduce**
1. Open chiaki-ng
2. Click on PS5 (unregistered) Automatic Registration Available
3. Click "Yes" on "Would you like to use automatic registration?"
4. chiaki-ng hangs
5. Force close the app

**Expected behavior**
chiaki-ng should connect to the PS5 as it had in the past. 

**Desktop (please complete the following information):**
 - OS: SteamOS
 - Device: Steam Deck
 - chiaki-ng Version 1.10.0

### LINKED ISSUE #800: [BUG] Remote Play via PSN fails with "SSL peer certificate ... was not OK" in Flatpak build, works fine in AppImage/native build
# [BUG] Remote Play via PSN fails with "SSL peer certificate ... was not OK" in Flatpak build, works fine in AppImage/native build

I am not a "network guy", so in case I misunderstand something or wrong in my reasoning and conclusions, please correct me.

> **Context:** This report is a deep-dive investigation and root-cause analysis for the reports previously opened in #798 and #799 (where identical symptoms were observed on Fedora/Void and SteamOS, but without complete diagnostic logs).

## TL;DR

The Flatpak build of chiaki-ng (`io.github.streetpea.Chiaki4deck`, runtime `org.kde.Platform//6.9`)
fails to establish a Remote Play connection **over PSN** (confirmed on Arch, Fedora and Void #798, SteamOS #799), works fine over
**local network**, other way get an infinite loading screen with logs:

```
[E] websocket_thread_func: Connecting to push notification WebSocket
wss://<SERVER_IP>-pushcl.np.communication.playstation.net/np/pushNotification
failed with CURL error SSL peer certificate or SSH remote key was not OK
[I] Canceling establishing connection over PSN
```

Root cause (best evidence available, see below): the PSN push-notification endpoint (`*.np.communication.playstation.net`, hosted on AWS, `44.x.x.x` range) serves only the **leaf certificate**, without the intermediate (`COMODO/Sectigo RSA Domain Validation Secure Server CA`). Verifying this chain requires the client to already trust that intermediate as an anchor.

On the host this can be fixed (`trust extract-compat` after adding the intermediate as an anchor), and this **is** correctly picked up by native tools (`curl`, `openssl s_client`) and by the same chiaki-ng version running as an **AppImage**. It is **not** picked up inside the Flatpak sandbox, even with `--filesystem=host-os:ro` already granted in the manifest. 

Direct inspection from inside the sandbox shows **both** conventional CA bundle paths (`/etc/ssl/certs/ca-certificates.crt` and `/etc/pki/tls/certs/ca-bundle.crt`)  exist but are **not** backed by the host's live trust store at all - both are owned  by `nfsnobody:nfsnobody` with an identical, stable timestamp across sandbox sessions run days apart, and neither contains the manually-added intermediate that is present and working on the host. 

This points to the runtime shadowing/embedding its own copies at both well-known paths regardless of `--filesystem=host-os:ro`, rather than a simple Debian-vs-Fedora path naming mismatch (see section 4 for the full finding - a naming mismatch was the initial hypothesis, subsequently narrowed down by direct inspection).

This is **not** a logic bug in chiaki-ng's own code - `holepunch.c` (the file that opens this WebSocket) never sets `CURLOPT_CAINFO`/`CURLOPT_CAPATH`, it relies entirely on the TLS backend's default resolution, which is where the Flatpak-vs-host mismatch causes the failure, as far as I understand.

---

## Environment

- **chiaki-ng version:** 1.10.0 (Flatpak build)
- **Install method:** Flatpak, `io.github.streetpea.Chiaki4deck`
- **Flatpak runtime:** `org.kde.Platform//6.9` (per `scripts/flatpak/io.github.streetpea.Chiaki4deck.yaml`)
- **Flatpak versions:** `1.18.0` and `1.18.1`
- **Host OS:** Arch Linux (Omarchy), kernel `7.1.8-arch1-3`
- **GPU / driver:** Intel UHD Graphics 630 (CFL GT2), Mesa 26.0.6, vaapi fallback (Vulkan video decode not supported by driver - unrelated to this issue, mentioned only for completeness)
- Connection tested from: mobile tethering, third-party Wi-Fi, and through a self-hosted AmneziaWG 2.0 (WireGuard-based) VPN - all reproduce the same failure. Ruled out network-level interference/DPI as the cause (see "Diagnostics already ruled out" below).

## Steps to Reproduce

1. Install `io.github.streetpea.Chiaki4deck` via Flatpak on a non-Fedora-based host (Arch confirmed; Debian/Ubuntu likely, see "Affected distros" below).
2. Register/pair a PS5 console for Remote Play via PSN (works fine locally on the same LAN).
3. Leave the home network (any external network - mobile data, other Wi-Fi, or through a VPN all reproduce it) and attempt "Remote Connection via PSN" to the same, already-registered console.
4. Observe: UI hangs on "establishing connection with console over PSN" indefinitely.

## Expected behavior

Remote Play connects over PSN the same way it does when running the same chiaki-ng version natively (AppImage) or when using the official PS Remote Play app.

## Actual behavior

Session log (trimmed, libplacebo surface-format enumeration removed for
brevity):

```
[I] Chiaki Version 1.10.0
[I] Trying to use hardware decoder "vaapi"
[I] Using hardware decoder "vaapi" with pix_fmt=vaapi
[I] ChiakiOpusEncoder initialized
[I] Duid: <redacted>
[I] Failed to find UPnP-capable devices on network: err=0
[E] websocket_thread_func: Connecting to push notification WebSocket
    wss://44-234-160-165-pushcl.np.communication.playstation.net/np/pushNotification
    failed with CURL error SSL peer certificate or SSH remote key was not OK
[I] Canceling establishing connection over PSN
```

This reproduced consistently across **multiple different `pushcl` edge IPs** resolved over several days/attempts (`44.234.160.165`, `44.229.19.180`, `44.233.39.168`, `44.234.161.133`, `44.224.29.103`), across three different networks, and both with and without a personal WireGuard-based VPN - ruling out a single flaky server node or a network-path issue.

---

### 1. The PSN push-notification endpoint serves an incomplete certificate chain

```
$ openssl s_client -connect 44.234.160.165:443 \
    -servername 44-234-160-165-pushcl.np.communication.playstation.net \
    -showcerts </dev/null
...
Certificate chain
 0 s:CN=*.np.communication.playstation.net
   i:C=GB, ST=Greater Manchester, L=Salford, O=COMODO CA Limited,
     CN=COMODO RSA Domain Validation Secure Server CA
...
Verify return code: 21 (unable to verify the first certificate)
```

Only the leaf certificate is sent - no intermediate. This is consistent across every edge IP/hostname observed above.

### 2. Fixing the host trust store fixes it for native tools, immediately

After adding the missing intermediate (`COMODO RSA Domain Validation Secure Server CA`, fetched from `crt.comodoca.com`) as a trust anchor on the host and running `sudo trust extract-compat`:

```
$ curl -Iv https://44-234-160-165-pushcl.np.communication.playstation.net/np/pushNotification
...
* SSL certificate verify ok.
...
< HTTP/1.1 404 Not Found          # (expected - HEAD isn't a valid method here, TLS is what matters)
```

```
$ openssl s_client -connect 44.233.39.168:443 \
    -servername 44-233-39-168-pushcl.np.communication.playstation.net -showcerts </dev/null
...
Verify return code: 0 (ok)
```

Confirmed reproducibly on **five separate edge IPs** over several days - every single one verifies cleanly once the host trust store has the intermediate.

### 3. The same fix is invisible inside the Flatpak sandbox

Tried, in order, all of the following, restarting the app between each:

- `flatpak override --user --env=SSL_CERT_FILE=<path to a merged bundle
  containing the fixed host ca-certificates.crt>`
- `flatpak override --user --env=CURL_CA_BUNDLE=<same path>`
- `flatpak override --user --filesystem=/etc/ssl:ro` (rejected outright:
  `F: Not sharing "/etc/ssl" with sandbox: Path "/etc" is reserved by
  Flatpak`)
- Full `flatpak override --user --reset` (falls back to whatever the
  manifest's own `--filesystem=host-os:ro` already grants)

**None of these changed the outcome.** The app still fails with the exact same `CURL error SSL peer certificate or SSH remote key was not OK`, on a freshly re-resolved edge IP each time, even though a manual `openssl s_client` run against that very IP, at that very moment, from the same machine, verifies cleanly.

Inspecting `lib/src/remote/holepunch.c` (the file that opens this specific WebSocket) confirms the app never calls `curl_easy_setopt(..., CURLOPT_CAINFO, ...)` or `CURLOPT_CAPATH` - it relies entirely on libcurl's default trustresolution. So this isn't something the app's own C code is overriding orgetting wrong; it's a property of how curl/OpenSSL resolve the default CA bundle *inside this specific Flatpak runtime*.

### 4. Confirmed mechanism: both conventional CA paths inside the sandbox are runtime-embedded, not host bind-mounts

The manifest (`scripts/flatpak/io.github.streetpea.Chiaki4deck.yaml`) uses:

```yaml
runtime: org.kde.Platform
runtime-version: '6.9'
finish-args:
  - --filesystem=host-os:ro
```

Initially assumed this was a simple "Debian-style vs. Fedora-style path" mismatch (host populates `/etc/ssl/certs/ca-certificates.crt`, runtime looks for `/etc/pki/tls/certs/ca-bundle.crt`, the two trees never overlap). Checked directly from inside the sandbox shell (`flatpak run --command=sh io.github.streetpea.Chiaki4deck`) - `strace` isn't available in this runtime (it ships with `org.kde.Platform`, not the `Sdk` variant), so this was verified by inspecting both files directly:

```
$ ls -la /etc/ssl/certs/ca-certificates.crt
lrwxrwxrwx 1 nfsnobody nfsnobody 20 Jun 18 19:42 /etc/ssl/certs/ca-certificates.crt -> ../tls-ca-bundle.pem
$ grep -c "COMODO RSA Domain Validation Secure Server CA" /etc/ssl/certs/ca-certificates.crt
0

$ ls -la /etc/pki/tls/certs/ca-bundle.crt
lrwxrwxrwx 1 nfsnobody nfsnobody 49 Jun 18 19:42 /etc/pki/tls/certs/ca-bundle.crt -> /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
$ grep -c "COMODO RSA Domain Validation Secure Server CA" /etc/pki/tls/certs/ca-bundle.crt
0
```

Both conventional paths **exist**, but:

- Neither contains the intermediate manually added to the host's trust store (confirmed present and working on the host itself via `openssl s_client`, see section 2 above).
- Both are owned by `nfsnobody:nfsnobody` with an identical timestamp (`Jun 18 19:42`) - this timestamp/ownership signature was observed identically across two separate sandbox sessions run days apart, which is inconsistent with a live bind-mount of the host's own (user-editable, Arch Linux) `/etc/ssl` tree, and consistent instead with both being read-only copies baked into the runtime's own OSTree image at build time.

This means the actual root cause is **not** a naming-convention mismatch between two host-vs-runtime path conventions - it's that **neither conventional CA path inside the sandbox is backed by the host's live trust store at all**, regardless of which naming convention you check. 

Despite `--filesystem=host-os:ro` being granted, the runtime appears to shadow/take precedence for these specific well-known certificate paths with its own bundled copies. This is plausibly a deliberate Flatpak/runtime design choice (security-critical trust files not depending on unpredictable host layouts), but as currently configured there does not appear to be a working, documented override path exposed to either the user (`SSL_CERT_FILE`/`CURL_CA_BUNDLE` env vars had no effect - consistent with this finding, since the app's libcurl very likely resolves its default CAINFO to one of these runtime-embedded files before anything env-based is ever consulted) or, apparently, to the app's own manifest via the currently granted permissions.

This class of problem (custom/added host CA certs invisible inside a Flatpak sandbox) is a known, long-standing, unresolved issue in the broader Flatpak ecosystem - see flatpak/flatpak#5967 and flatpak/flatpak#4603 - but this report is, as far as I could find, the first documentation of it specifically in chiaki-ng's PSN Remote Play path, where it matters because Sony's own server-side chain is *also*, it seems, incomplete, which is what exposes the gap (a server with a complete chain never triggers this).

### 5. Definitive isolation: same version, no sandbox, works immediately

Downloaded the official AppImage release (chiaki-ng 1.10.0, same commit), imported settings/console pairing, attempted the identical PSN Remote Connection from the same machine, same network, same moment:

```
[I] Chiaki Version 1.10.0
...
[I] >> Created session
[I] holepunch_session_create_offer: IPV6 NOT supported by your PlayStation
    console. Skipping IPV6 connection
[I] >> Created offer msg for ctrl connection
[I] >> Started session
[I] Sent response to <redacted-ipv4>:21803
[I] >> Punched hole for control connection!
[I] REGIST - Starting RUDP session
[I] xvnvdu-ps5 successfully registered for Remote Play
[I] Ctrl connected
[I] Ctrl received Login message: success
[I] Punching hole for data connection
```

Connects and streams successfully on the first attempt. This isolates the problem entirely to the Flatpak sandbox/runtime, not to chiaki-ng's own connection logic, network conditions, or console/account state.

---

## Open question: why did this work before, and why now?

This setup (same Flatpak build, same host, same console) was working fine over PSN until fairly recently, then started failing without any change on my end other than re-pairing the console and updating the PS system software around the same time.

Given the finding above - that the sandbox's CA trust is a fixed snapshot baked in at runtime build time, not something that changes based on the host

> The sandbox side of things should have behaved identically the whole time. That leaves the server side as the most likely variable that actually changed: **my best guess is that the PSN push-notification endpoint used to serve a complete certificate chain (leaf + intermediate) and, at some point, stopped including the intermediate** - possibly due to a routine certificate rotation or a load-balancer/CDN config change on Sony's side (note the current leaf certificate's `NotBefore: Mar 4 2026`, i.e. it's only a few months old at the time of this report, consistent with a recent rotation).
This would explain why verification used to succeed without any of the workarounds described in this issue, and why it no longer does, without requiring anything to have changed in chiaki-ng, Flatpak, or on my host.

**I want to be upfront that this is a plausible reconstruction, not a confirmed fact** - I have no way to check what the server was actually sending a month or two ago. The re-pairing/PSN update happening around the same time is very likely coincidental (a change on Sony's global infrastructure has no reason to correlate with one user's console maintenance), but I can't rule out some other cause specific to my own setup either. 

If anyone else can confirm whether this endpoint's certificate chain has changed recently, or if the official PS Remote Play app is unaffected by this (which would help distinguish "Sony's server changed" from "chiaki-ng's TLS stack is pickier than the official client's"), that would help pin this down further.

---

## Diagnostics already ruled out

- **Network-level interference / DPI / MITM proxy:** tested across three distinct networks (mobile tethering, third-party Wi-Fi, home) and through a personal WireGuard-based VPN with a dedicated remote server - the exact same failure reproduces every time on the Flatpak build, and the exact same success reproduces every time on the AppImage build. `openssl s_client` through the VPN, direct to the same edge IP chiaki-ng was using at that moment, verified cleanly against the (fixed) host trust store.
- **Flaky/misconfigured single PSN edge server:** reproduced across 5+ distinct edge IPs over several days.
- **Stale/incomplete host trust store:** ruled out - confirmed fixed and stable via `openssl s_client`/`curl` on the host, repeatedly, over multiple sessions.
- **Console/pairing/account state:** ruled out by the AppImage test using the exact same imported pairing data succeeding immediately.

---

## Affected distros (best guess, not exhaustively tested)

Since neither conventional CA path inside the sandbox appears to be backed by the host's live trust store (see section 4), this most likely affects **any** host distro equally, including Fedora-family ones - not just Debian/Arch-style hosts as initially suspected. The gap only becomes visible when:

1. A server's certificate chain is incomplete (so verification actually depends on a manually-trusted anchor rather than the runtime's bundled default Mozilla CA set), and
2. I fixed this locally on the host and reasonably expect that fix to apply everywhere, including inside sandboxed apps.

Confirmed affected: Arch Linux, Fedora/Void (#798), SteamOS (#799). Not yet tested on other host distros - if anyone reading this can confirm/deny on other hosts, that would help narrow this down further, but it seems like  at this point this does not depend on a certain distro.

---

## Possible directions for a fix (not proposing one specific approach yet)

A few options, each with different tradeoffs - happy to attempt a PR once there's agreement on preferred direction:

1. Add an explicit bind of the Fedora-style CA path from the host into the manifest's `finish-args` (only helps hosts that happen to also populate that path, e.g. Fedora - doesn't fix the mismatch for Debian/Arch hosts).

2. A small wrapper/init step at app startup that looks for a host CA bundle at a few well-known locations (`/etc/ssl/certs/ca-certificates.crt`, `/etc/pki/tls/certs/ca-bundle.crt`, etc., via whatever host paths `host-os:ro` exposes) and symlinks/copies it to wherever the bundled curl/OpenSSL actually resolves its default from.

3. Escalate to upstream Flatpak/freedesktop-sdk, referencing #5967/#4603, since this is fundamentally the same unresolved platform-level gap.

4. **Application-level trust:** If the PSN push notification endpoint intermediate CA is stable/known, chiaki-ng could bundle the required Sectigo/COMODO intermediate cert directly within the package and load it via `CURLOPT_CAINFO` / `CURLOPT_CAPATH` or OpenSSL context initialization specifically for this connection.

I'm not a C/C++ developer, but I'm willing to test any proposed patch against this exact reproducible scenario, and can also try to put together a PR if a maintainer can point me to the right place in the build/manifest to hook it in.

---

## Related upstream issues

- flatpak/flatpak#5967 - "Unable to load host certificates"
- flatpak/flatpak#4603 - "Some flatpak applications cannot access valid certs"

Both describe the same general class of problem (custom-added host CA certs invisible inside Flatpak sandboxes), unresolved as of this writing, as far as I understand.

xlanor:
As discussed in Discord:

psn's mobile push notif servers returns a node to use, and some of these nodes only serve a leaf-only chain without the necessary intermediary. If the TLS stack doesn't implement AIA, this will cause the verification to fail.

This adds AIA chain recovery for basic repairing of the cert chain without needing to ship certificates we don't own to end users.

Opted for this approach:

- Let curl fail
- confirm it failed on the cert, not a dns or timeout
- read the leaf served, repeat until path reaches root
- retry the recovered intermediates with `CURLOPT_CAINFO_BLOB`
- cache the bundle

Caveats:

- no full backtracking here. Only took the first `caIssuers` URL and checked it, capped at 4 hops
- At this point of time DER or PEM are the only ones supported, so cross-signed CAs will fail.

 If we run into this issue again I think it should be simple to update the implementation to handle this, but for now a single hop is sufficient to recover the intermediate certificate.

`aia.h` declares six certificate functions that `aia.c` calls but does not implement. in calling order:

```text
chiaki_aia_peek_leaf         read what the host served, unverified
chiaki_aia_cert_describe     log the leaf we are walking from
  chiaki_aia_path_completes  leaf + what we have -> a trusted anchor?
  chiaki_aia_issuer_url      if not, the caIssuers URL off the cert we are standing on
  (fetch)                    a function pointer, not part of this
  chiaki_aia_pem_to_der      normalise the answer, DER or PEM
  chiaki_aia_cert_describe   log the hop, then round again
chiaki_aia_der_to_pem        once it completes, bundle what we collected for CURLOPT_CAINFO_BLOB
```

the indented four are the loop, capped at four hops.

This will allow different backends to be implemented by any forks which may vary from platform to platform. For NX, openSSL doesn't exist and libNX doesn't provide any X509 validation functions so I have to bring back mbedTLS for this one. I know theres a webOS fork floating around somewhere and I have zero clue what they use, so hopefully this allows them to implement it their way.

OpenSSL backend is implemented here

I've also added a public API to allow you to pin the node to a specific one for easy reproduction. 

Logs are from my end (akira) with mbedTLS backend to verify the certs that are walked, you'll likely need to do more testing on your side for linux but let me know so that we can stay on the same interfaces and make it easy for me to rebase and push any changes to you.

```
04:59:53.397[INFO] Connect to PS5-<REDACTED> (Remote)
04:59:53.410[INFO] Switched to connection log: sdmc:/switch/akira/logs/230826_045953_remote.log
04:59:53.410[INFO] ========================================
04:59:53.410[INFO] CONNECTION ATTEMPT: remote to PS5-<REDACTED> (Remote)
04:59:53.410[INFO] ========================================
04:59:53.410[INFO] Connection thread started
04:59:53.410[INFO] Initiating holepunch connection...
04:59:53.412[INFO] Holepunch session state: 1 = [ ✅INIT ]
04:59:53.413[INFO] Port guessing enabled: 75 guesses, 120 sockets
04:59:53.413[WARNING] Forcing push notification node to 44-233-186-110-pushcl.np.communication.playstation.net
04:59:53.413[INFO] Holepunch session initialized for PS5-<REDACTED> (Remote)
04:59:53.413[INFO] Starting holepunch connection sequence for PS5-<REDACTED> (Remote) (PS5) - attempt 1/1
05:00:00.413[WARNING] UPnP discovery timed out after 7000 ms, skipping
05:00:00.413[INFO] UPNP discovery completed
05:00:00.413[INFO] Creating holepunch session on PSN...
05:00:00.413[WARNING] chiaki_holepunch_session_create: using overridden websocket FQDN 44-233-186-110-pushcl.np.communication.playstation.net
05:00:00.413[INFO] chiaki_holepunch_session_create: Created websocket thread
05:00:00.413[INFO] websocket_thread_func: connecting to 44-233-186-110-pushcl.np.communication.playstation.net with peer verification ENABLED (aia blob 0 bytes, gen 0)
05:00:00.413[INFO] chiaki_holepunch_session_create: Waiting for websocket to open...
05:00:01.356[INFO] aia: read a 1742 byte certificate from 44-233-186-110-pushcl.np.communication.playstation.net
05:00:01.356[INFO] aia: walking from CN=*.np.communication.playstation.net (issued by C=GB, ST=Greater Manchester, L=Salford, O=COMODO CA Limited, CN=COMODO RSA Domain Validation Secure Server CA)
05:00:01.358[INFO] certchain: 54 built-in root(s) loaded, 0 skipped (63 entries, direct)
05:00:01.358[INFO] aia: path incomplete (flags 0x8): The certificate is not correctly signed by the trusted CA

05:00:01.358[INFO] aia: fetching issuer 1 from http://crt.comodoca.com/COMODORSADomainValidationSecureServerCA.crt
05:00:01.379[INFO] aia:   hop 1 -> C=GB, ST=Greater Manchester, L=Salford, O=COMODO CA Limited, CN=COMODO RSA Domain Validation Secure Server CA (issued by C=GB, ST=Greater Manchester, L=Salford, O=COMODO CA Limited, CN=COMODO RSA Certification Authority)
05:00:01.381[INFO] certchain: 54 built-in root(s) loaded, 0 skipped (63 entries, direct)
05:00:01.382[INFO] aia: path complete after 1 hop(s), 2151 byte bundle
05:00:01.382[INFO] websocket_thread_func: connecting to 44-233-186-110-pushcl.np.communication.playstation.net with peer verification ENABLED (aia blob 2151 bytes, gen 1)
05:00:01.418[INFO] Failed to find UPnP-capable devices on network: err=0
05:00:01.418[INFO] chiaki_holepunch_session_create: Waiting for websocket to open...
05:00:02.000[INFO] websocket_thread_func: Connected to push notification WebSocket wss://44-233-186-110-pushcl.np.communication.playstation.net/np/pushNotification
05:00:02.000[INFO] Holepunch session state: 3 = [ ✅INIT ✅WS_OPEN ]
05:00:02.000[INFO] http_create_session: Sending JSON:
{"remotePlaySessions":[{"members":[{"accountId":"me","deviceUniqueId":"me","platform":"me","pushContexts":[{"pushContextId":"<UUID>"}]}]}]}
05:00:02.000[INFO] websocket_thread_func: PING.
05:00:02.148[INFO] websocket_thread_func: Received WebSocket frame of length 0 with flags 64
05:00:02.148[INFO] websocket_thread_func: Received PONG.
05:00:02.303[INFO] websocket_thread_func: Received WebSocket frame of length 733 with flags 1
05:00:02.303[INFO] websocket_thread_func: Received WebSocket frame with 733 bytes of payload.
05:00:02.303[INFO] {
  "version":"2.1",
  "method":3001,
  "dataType":"psn:sessionManager:sys:rps:members:created",
```

#798 , #800


xlanor:
Darwin failures unrelated, due to ffmpeg

xlanor:
Just an aside, but I think Sony might have taken the misconfigured nodes out of service.

The test servers I had dont work anymore

